Issue #6, opened 2017-08-14
Increase / Decrease should not have marked value.


amgrubb:
The satisfaction label box should not appear for Increase or Decrease (the bottom on in the screen shot). The marked value is FS and FD respectively for passing to the backend. 
![screen shot 2017-08-14 at 1 40 31 pm](https://user-images.githubusercontent.com/6596937/29285556-de1cb75a-80fc-11e7-8d49-0f8b831938da.png)

Akira83:
If I understood correctly it should be like this:
![increase](https://user-images.githubusercontent.com/5030821/29374891-09dab452-8281-11e7-8c1f-0f3a3ee0bd14.jpg)
![decrease](https://user-images.githubusercontent.com/5030821/29374896-0cb465f6-8281-11e7-9e24-93e85bece695.jpg)

amgrubb:
Yes that is correct.

amgrubb:
Updated Online

amgrubb:
I screwed up, this isn't actually a bug and was correct in the first place.
Can we undo the pull request?

amgrubb:
Some of this code was useful, so we need to figure out what to keep.

amgrubb:
I reverted the change, and now none of the market values are saving properly.
I don't see how any of these changes affect that.

Akira83:
I will close this issue. If the problem reaper we open it again.

Issue #5, opened 2017-08-14
Dynamic Function Marked Value doesn't save.


amgrubb:
Done of the marked values (the satisfaction value used in the dynamic function) are saving when you select them in the interface. They are returning to the default.

This includes: MP, MN, and RC.
![screen shot 2017-08-14 at 1 36 45 pm](https://user-images.githubusercontent.com/6596937/29285442-84a8402c-80fc-11e7-93bf-7b6ce32cc47f.png)

This results in incorrect or missing marked values in the backend.

Akira83:
Fixed in frontend, not verify if the correct value is going to backend. This one was a bit hard to fix, it took more time to find where change the code.

Akira83:
I will not create a pull request for this because when I tried to create a new pull request it creates with the previous commits that I did. So I will wait until the previous pull request is finished. Then I submit this new one.

amgrubb:
Previous pull request was approved.

Akira83:
I will close this issue, if the problem reaper we open it again.

PR #25, opened 2017-08-21
Marcel dev

Akira83:
I executed some tests and I think that with this modifications I was able to solve issues #6 and #5 .
More tests are required to see if its behaviour is correct.

### DIFF
--- a/leaf-ui/scripts/main.js
+++ b/leaf-ui/scripts/main.js
@@ -4,9 +4,6 @@ var paper;
 var stencil;
 var mode;
 
-// Mode specific variables
-var graphObject = new graphObject();		// Stores all variables between modes
-
 var linkInspector = new LinkInspector();
 var elementInspector = new ElementInspector();
 var constrainsInspector = new ConstraintsInspector();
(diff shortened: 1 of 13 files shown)


leaf-ui/scripts/main.js
```diff
@@ -26,7 +25,10 @@ var reader;
 
 //Properties for both core and simulator.
 //TODO: merge this two arrays in order to make use the same name for all
-var satvalues = {satisfied: 2, partiallysatisfied: 1, partiallydenied: -1, denied: -2, unknown: 4, conflict: 3, none: 0};
+var satvalues = {
```
amgrubb: Why are we still using the iStar values here. How are the four conflict types stored?

leaf-ui/scripts/main.js
```diff
@@ -47,6 +49,14 @@ mode = "Modelling";		// 'Analysis' or 'Modelling'
 linkMode = "Relationships";	// 'Relationships' or 'Constraints'
 
 graph = new joint.dia.Graph();
```
amgrubb: Why was their a graphObject and a graph before but now we only need a graph?
Does this any anything to do with the constraint view?

leaf-ui/scripts/main.js
Akira83: That is why I put that "//TODO" comment, a lot of parts in the code for example: the select initial values are using one type of annotation in the satvalues array and some parts of the code are using the other array with "0000, 0001, etc..." type of annotation. Since they "mean" the same thing it is a good idea refactor and use just one.

leaf-ui/scripts/main.js
Akira83: This one I did when I was refactoring the code. If you look the graphObject it has the same information that we are able to get from the graph (joint.dia.Graph) and in the end we were having to make updates every time the graph object changed. Since the old code was getting information from the graphObject it was getting values not up to date, and as in javascript objects you can add attributes dinamically, we can do that direct in the graph object as it is being done in a lot of parts of the code.

Issue #23, opened 2017-08-17
Increase / Decrease Bug


amgrubb:
- [x] All none as initial value for increase/decrease functions.
Increase with an initial value none is valid input.
![screen shot 2017-08-17 at 11 06 06 am](https://user-images.githubusercontent.com/6596937/29419071-20209702-833c-11e7-9dbf-d474c48a64d0.png)
But backend outputs this error:
`INC Dynamic Value for intention 0002 has Unknown/None/Conflict value.`

amgrubb:
- [ ] With Strong conflict prevents, (PS,PD) etc. causes incorrect values.

- [ ] Same change will be required in UD functions.

Testing:
- [ ] Test INC / DEC
- [ ] Test MP / MN
- [ ] Test UD Function

amgrubb:
Not necessarily fixed, but this bug is out of date.

Issue #19, opened 2017-08-17
Error Passing to User: Over-constrained model


amgrubb:
When the CSP solver discovers that the model is over-constrained it outputs:

`Constraint: And733 : And(Or2089 : Or( XeqC4971 : XeqC(N0046_0_PS=1, 0 ), XeqC4972 : XeqC(N0038_0_PS=0, 1 )),Or2090 : Or( XeqC4973 : XeqC(N0046_0_PD=0, 0 ), XeqC4974 : XeqC(N0038_0_PD::{0..1}, 1 )),,
have conflicting constraints, not solvable
Found Solution = False`

This should be passed to the front-end, parsed, and communicated to the user.

g2hchang:
Incorporated into #95.

PR #26, opened 2017-08-21
Updated Increasing and Decreasing Functions for Path Constraints

amgrubb:
#23 Fixed
Also started to work on #19 but not finished yet.


Akira83:
Hello Alicia, I test this model and didn't work. Is the frontend sending something wrong to the backend? Let me know so I can fix it.
increasetest

[increaseTest.json.txt](https://github.com/amgrubb/BloomingLeaf/files/1242159/increaseTest.json.txt)

amgrubb:
Ok, looking into it.

amgrubb:
It doesn't appear to have anything to do with the increase/decrease changes.

amgrubb:
One error in the file is it says:
    "currentState": "",
should be:
    "currentState": "0",

amgrubb:
These two elements are new:
    "initialAssignedEpoch": "",
    "initialValueTimePoints": "",
What are they?

amgrubb:
The error occurs on TroposCSPAlgorithm line ~118
`if (this.spec.getInitialValueTimePoints().length != this.spec.getInitialValues()[0].length)`
this.spec.getInitialValues() returns null, so the initial values are not being set.

This is a new / different bug.

Issue #45, opened 2017-08-27
UD won't allow Decrease (D) to have value.


amgrubb:
When I create the following function. I can't select anything in the box next to Decrease. I should be able to select an evaluation label.

![screen shot 2017-08-27 at 11 19 01 am](https://user-images.githubusercontent.com/6596937/29751501-4605f5d8-8b1c-11e7-86d3-8c75f7c5073c.png)

g2hchang:
Created a pull request #54

PR #54, opened 2017-09-09
Decrease bug fix

g2hchang:
Fixed the bug #45 by changing data-type of values defined in satvalues in main.js. Javascript has unexpected results when comparing string numbers.

### DIFF
--- a/leaf-ui/scripts/main.js
+++ b/leaf-ui/scripts/main.js
@@ -29,7 +29,7 @@ var reader;
 //TODO: merge this two arrays in order to make use the same name for all
 var satvalues = {
 		"satisfied": 2, "partiallysatisfied": 1, "partiallydenied": -1, "denied": -2, "unknown": 4, "conflict":3, "none": 0,
-		2: "satisfied", 1: "partiallysatisfied", "-1": "partiallydenied", "-2": "denied", "4": "unknown", "3": "conflict", "0": "none"
+		"2": "satisfied", "1": "partiallysatisfied", "-1": "partiallydenied", "-2": "denied", "4": "unknown", "3": "conflict", "0": "none"
 		};
 
 var satValueDict = {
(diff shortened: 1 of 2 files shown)


amgrubb:
Feature pull requests should use `develop` has their base, not `master`.

leaf-ui/scripts/main.js
```diff
@@ -28,7 +28,7 @@ var reader;
 //Properties for both core and simulator.
 //TODO: merge this two arrays in order to make use the same name for all
 var satvalues = {
-		"satisfied": "2", "partiallysatisfied": "1", "partiallydenied": "-1", "denied": "-2", "unknown": "4", "conflict": "3", "none": "0",
+		"satisfied": 2, "partiallysatisfied": 1, "partiallydenied": -1, "denied": -2, "unknown": 4, "conflict":3, "none": 0,
```
amgrubb: This seems to be the only actual change in this pull request?

amgrubb:
Do you know why this was only an issue for decrease? Doesn't seem like this should be the full fix.

amgrubb:
#45 was affected by how `ElementInspector.js` lines ~ 210-231 read the values of `satvalues`.

```javascript
this.chartHTML.positiveOnly = function(currentVal){
      currentVal = satvalues[currentVal];
      result = '';
      for (var i = currentVal; i <= satvalues['satisfied']; i++){
        // Find text by value eg. given i = -1, we want to find partiallydenied as satvalues[partiallydenied] = -1
        //var text = Object.keys(satvalues).find(key => satvalues[key] === i);
        var text = satvalues[i];
        result += eval(text);
      }
      return result;
    }
    this.chartHTML.negativeOnly = function(currentVal){
      currentVal = satvalues[currentVal];
      result = '';
      for (var i = currentVal; i >= satvalues['denied']; i--){
        // Find text by value eg. given i = -1, we want to find partiallydenied as satvalues[partiallydenied] = -1
        //var text = Object.keys(satvalues).find(key => satvalues[key] === i);
    	  var text = satvalues[i];
    	  result += eval(text);
      }
      return result;
    }
```

amgrubb (approved):

Issue #71, opened 2017-09-19
Clear old analysis data when switching between Modeling and Analysis


amgrubb:
"currentState":"5|5","initialAssignedEpoch":"TA0_0,E0001_74,TA2_2,TA1_1,TA4_4,TE8_74,E0000_72,TA3_3,TE7_72,TA6_6,TA5_5"

Akira83:
Working on local. Need testing on server.

Akira83:
Was cleaning data in incorrect place. Corrected that, it cleans data when switch mode Analysis-> modeling

Akira83:
Tested on Server and it erases the old analysis data when the user changes from analysis to modeling.

PR #73, opened 2017-09-20
Feature clean old analysis

Akira83:
Cleaning the old analysis

BUG #71 - Tested in my server seems to be working properly. Cleans old data when switch mode Analysis-> modeling.

### DIFF
--- a/leaf-ui/js/object/InputAnalysis.js
+++ b/leaf-ui/js/object/InputAnalysis.js
@@ -21,12 +21,22 @@ function getAnalysisValues(analysisInterface){
 	analysisInterface.elementList = getElementList();
 	
 	//Data required for 2. Explore Possible Next States
+	//Erase old initialAssignedEpoch data
+	if(analysisInterface.initialAssignedEpoch.length > 0){
+		analysisInterface.initialAssignedEpoch = "";
+	}
 	if($('#finalAssigneEpoch').val()){
 		analysisInterface.initialAssignedEpoch = $('#finalAssigneEpoch').val();		
+	}
+	
+	//Erase old initialAssignedEpoch data
+	if(analysisInterface.initialValueTimePoints.length > 0){
+		analysisInterface.initialValueTimePoints = "";
 	}
 	if($('#finalValueTimePoints').val()){
 		analysisInterface.initialValueTimePoints = $('#finalValueTimePoints').val();		
 	}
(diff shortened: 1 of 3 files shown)


leaf-ui/rappid-extensions/AnalysisInspector.js
```diff
@@ -77,7 +77,6 @@ var AnalysisInspector = Backbone.View.extend({
 		// These functions are used to communicate between analysisInspector and Main.js
 		this._analysisFunctions = analysisFunctions;
 		this.$el.html(_.template(this.template)());
-		$('head').append('<script src="./scripts/js-objects/analysis.js"></script>');
```
amgrubb: Why is this script no longer needed?
Is it called from somewhere else?

leaf-ui/js/object/InputAnalysis.js
```diff
@@ -21,12 +21,16 @@ function getAnalysisValues(analysisInterface){
 	analysisInterface.elementList = getElementList();
 	
 	//Data required for 2. Explore Possible Next States
+	//Erase old initialAssignedEpoch data
```
amgrubb: This comments are no accurate anymore?

leaf-ui/js/object/InputAnalysis.js
```diff
@@ -21,12 +21,16 @@ function getAnalysisValues(analysisInterface){
...
 	if($('#finalAssigneEpoch').val()){
 		analysisInterface.initialAssignedEpoch = $('#finalAssigneEpoch').val();		
 	}
+	
+	//Erase old initialAssignedEpoch data
```
amgrubb: This comment is not accurate anymore?

leaf-ui/rappid-extensions/AnalysisInspector.js
Akira83: This script is still needed. It is called on analysis.html that pops up for "get next possible states" analysis. I had to remove from there because for some reason it started to crash when I tried to execute locally. But I need to figure out how to add it again some how without crashing...
I will fix this before making the pull request again

leaf-ui/js/object/InputAnalysis.js
Akira83: True... don't know why I left there... will erase.

PR #77, opened 2017-09-22
Feature clean old analysis

Akira83:
Comments on changes:
- Created a new JS var to hold the necessary data for following analysis.
- Removed the html elements that were used to save these data.
- Clear the value every time the change between modes is called, so every time the user changes for modeling or analysis view the variables are cleaned.

### DIFF
--- a/leaf-ui/rappid-extensions/AnalysisInspector.js
+++ b/leaf-ui/rappid-extensions/AnalysisInspector.js
@@ -77,6 +77,7 @@ var AnalysisInspector = Backbone.View.extend({
 		// These functions are used to communicate between analysisInspector and Main.js
 		this._analysisFunctions = analysisFunctions;
 		this.$el.html(_.template(this.template)());
+		$('head').append('<script src="./scripts/js-objects/analysis.js"></script>');
 	},
 	// Function called by Simulate Single Path.
 	singlePath: function(){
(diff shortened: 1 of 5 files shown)


leaf-ui/scripts/main.js
```diff
@@ -257,10 +257,6 @@ $('#analysis-btn').on('click', function(){
 //Switch to modeling mode
 $('#model-cur-btn').on('click', function(){
 	switchToModellingMode(false);
-});
-
-$('#model-init-btn').on('click', function(){
```
Akira83: This is a dead code. In the current project we don't have any html element with this id.

leaf-ui/rappid-extensions/AnalysisInspector.js
```diff
@@ -77,6 +77,7 @@ var AnalysisInspector = Backbone.View.extend({
 		// These functions are used to communicate between analysisInspector and Main.js
 		this._analysisFunctions = analysisFunctions;
 		this.$el.html(_.template(this.template)());
+		$('head').append('<script src="./scripts/js-objects/analysis.js"></script>');
```
Akira83: Added back this line so analysis window pops up. Removed before because this line was crashing the code and I couldn't even test the changes.

Issue #61, opened 2017-09-15
Only return single state.


amgrubb:
Change the length of the all paths analysis to only return the values for the new state.

Issue #32, opened 2017-08-24
All the ways to call the back-end for analysis.


amgrubb:
Types of Analysis:

The front-end is supposed to call the backend in the following configurations and results.
- [x] Single path from time point 0, without other time points. -> Produces forward analysis.
- [x] Single path from time point 0, with time points from dynamic functions or user indicated absolute or relative time points. -> Give single path simulation result.
- [ ] Single path from time point X. -> Pass initial values from 0-X to back-end, and it will give single path simulation result from 0-X-end.
- [ ] All next states from time point X. -> Pass initial values from 0-X and backend will return all possible next states.
- [ ] All next states from time point 0. -> Pass initial values 0 and backend will return all possible next states. Solver will generate unassigned values for None with NT function type.

At the moment only the second one sort of works.

amgrubb:
- [ ] 

 Single path from time point 0, without other time points. -> Produces forward analysis.
 Single path from time point 0, with time points from dynamic functions or user indicated absolute or relative time points. -> Give single path simulation result.

amgrubb:
> All next states from time point 0. -> Pass initial values 0 and backend will return all possible next states. Solver will generate unassigned values for None with NT function type.

Is actually two cases:

- [ ] At time point 0 with assigned initial values.
- [ ] At time point 0 with out assigned initial values.

PR #79, opened 2017-09-22
Get all states analysis working on backend.

amgrubb:
Makes process on #32 and #61.

See commit comments.


leaf-ui/scripts/main.js
```diff
@@ -26,7 +28,7 @@ var reader;
 //Properties for both core and simulator.
 //TODO: merge this two arrays in order to make use the same name for all
 var satvalues = {
-		"satisfied": "2", "partiallysatisfied": "1", "partiallydenied": "-1", "denied": "-2", "unknown": "4", "conflict": "3", "none": "0",
+		"satisfied": 2, "partiallysatisfied": 1, "partiallydenied": -1, "denied": -2, "unknown": 4, "conflict":3, "none": 0,
```
amgrubb: We might need two of these. One with quotes and one without. @g2hchang added the quotes to fix a different bug.

Issue #85, opened 2017-09-27
[BUG] ModelSpecBuider


Akira83:
I'm getting an error in in ModelSpecBuilder line 171. The problem seems to be related to the dynamic without MarkedValue. Not sure if the dynamic type NT should have some marked value, or should not be added as a dynamic. 

Solutions that I have in mind:
frontend send as marked value: 0000 when NT or backend add 0000 as marked value if it is correct.

Test case:
```{
  "analysis": {
    "action": "allNextStates",
    "maxAbsTime": "100",
    "conflictLevel": "S",
    "numRelTime": "5",
    "absTimePts": "1 3",
    "currentState": "5|55",
    "initialAssignedEpoch": [
      "TA0_0",
      "TA2_3",
      "TR3_90",
      "TA1_1",
      "TR5_22",
      "TR4_82",
      "TR7_55",
      "TR6_50"
    ],
    "initialValueTimePoints": [
      "0",
      "1",
      "3",
      "22",
      "50",
      "55",
      "82",
      "90"
    ],
    "elementList": [
      {
        "id": "0001",
        "status": [
          "0100",
          "1110",
          "0100",
          "0000",
          "1110"
        ]
      },
      {
        "id": "0001",
        "status": [
          "0100",
          "1110",
          "0100",
          "0000",
          "1110"
        ]
      }
    ]
  },
  "model": {
    "actors": [
      
    ],
    "intentions": [
      {
        "nodeActorID": "-",
        "nodeID": "0000",
        "nodeType": "G",
        "nodeName": "Goal_0"
      },
      {
        "nodeActorID": "-",
        "nodeID": "0001",
        "nodeType": "G",
        "nodeName": "Goal_1"
      }
    ],
    "links": [
      {
        "linkType": "AND",
        "linkSrcID": "0001",
        "linkDestID": "0000",
        "postType": null,
        "markedValue": null
      }
    ],
    "dynamics": [
      {
        "intentionID": "0000",
        "dynamicType": "NT",
        "line": null
      },
      {
        "intentionID": "0001",
        "dynamicType": "NT",
        "line": null
      }
    ],
    "constraints": [
      
    ]
  }
}

Akira83:
This is strange, because in the single path the marked value is being send correct.

amgrubb:
All dynamics and "NT" should have a marked value.
Use "0000" when it is not relevant.

PR #92, opened 2017-10-11
Hotfix marked value

Akira83:
Correction of issue #85 the marked value being passed was integer instead string.


leaf-ui/js/object/InputDynamic.js
```diff
@@ -29,6 +29,8 @@ function getDynamics(){
 	    
 	    if (isNaN(parseInt(initValue))){
 			initValue = satValueDict[initValue];		 
+		}else{
```
Akira83: The inital value (in this case marked value) was set as integer while modelspecbuilder class was expecting a string with 4 chars.

Issue #98, opened 2017-10-18
Graph update for increase/decrease


Akira83:
The graph is still not updating correctly around the increase/decrease bug.

amgrubb:
Nice! I like this fix.

PR #99, opened 2017-10-18
Hotfix for increase decrease function

Akira83:
It seems to be working, however, I think that it can be improved.

For instance, when PS the marked value should be only FS for Increase.
Right now it is PS and FS but if you select PS there is no increase
function.

Pull related to:
#98

### DIFF
--- a/leaf-ui/rappid-extensions/ElementInspector.js
+++ b/leaf-ui/rappid-extensions/ElementInspector.js
@@ -346,7 +346,7 @@ var ElementInspector = Backbone.View.extend({
     else if ($.inArray(functionType, funct_with_sat_value) > -1){
 	 //Clear marked Values
 	  var cell = this._cellView.model;
-	  cell.attr(".constraints/markedvalue", "0");
+	  delete cell.attr(".constraints/markedvalue");
 	  
   	  this.showFunctionSatValue(null);
       this.$('#user-constraints').hide();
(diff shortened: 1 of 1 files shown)


leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -346,7 +346,7 @@ var ElementInspector = Backbone.View.extend({
     else if ($.inArray(functionType, funct_with_sat_value) > -1){
 	 //Clear marked Values
 	  var cell = this._cellView.model;
-	  cell.attr(".constraints/markedvalue", "0");
+	  delete cell.attr(".constraints/markedvalue");
```
Akira83: Cleaning the markedvalue when not necessary.

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -508,15 +508,11 @@ var ElementInspector = Backbone.View.extend({
     var cell = this._cellView.model;
     var text = this.$('.function-type').val();
     var initVal = satvalues[this.$('#init-sat-value').val()];
-    var val;
+    var val = satvalues[this.$('#markedValue').val()];
```
Akira83: Get the marked value from frontend.

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -508,15 +508,11 @@ var ElementInspector = Backbone.View.extend({
...
+    
     if(cell.attributes.attrs['.constraints']){
-        val = cell.attributes.attrs['.constraints'].markedvalue;
```
Akira83: Removing the pre set marked value.

Issue #100, opened 2017-10-18
Link and the dynamics associated 


Akira83:
When you delete a link, the dynamics associated with the nodes connected to the link are deleted.

amgrubb:
@Akira83 This is still not fixed. When you change the link time, the dynamic function are still removed.

PR #106, opened 2017-11-01
Hotfix Issue #100

Akira83:
Issue #100:
Everytime you change the link type all tags from nodes are deleted.

Correction: Error in the code for when the tag was not NB.


leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -171,13 +170,22 @@ var LinkInspector = Backbone.View.extend({
     		  return false;
     	  }
 
+    	  //Verify if the node has the NB tag
```
Akira83: Added a function to verify if the function tag in NB.

PR #107, opened 2017-11-02
Bug 97

smfsamir:
The issue was that the init-sat-value <select> element has values of ["none", "satisfied", ...], i.e they're in words rather than bits. However, the satisfaction values of elements are stored as bits so this conversion was not being made properly when the elements were rendered. 

Furthermore, the commenting-out of line 1035 is a temporary fix to prevent the function type from being reset; it looks like the handler function for a change to "function-type" isn't quite right but we'll have to discuss what we want this to do.


g2hchang:
Apparently, the fix seems to introduce another issue.
With the fix, when I loads and try to select Initial Satisfaction Value, the dropdown box is not selectable.
screen shot 2017-11-02 at 6 02 54 pm

I believe the correct behavior should be to make the value selectable : 
screen shot 2017-11-02 at 6 02 33 pm 
**This is what I see in develop branch**
I am tagging @amgrubb to double check with her.

Issue #108, opened 2017-11-02
Executing for second time


Akira83:
Reproduce bug:
1 - Create a model and execute.
2 - Create a new model and execute.
BUG: The element list is filled wrong.

PR #109, opened 2017-11-02
Executing for second time

Akira83:
BUG #108
The element list was being keep from the old execution.


leaf-ui/scripts/main.js
```diff
@@ -280,6 +280,7 @@ $('#analysis-btn').on('click', function(){
 $('#model-cur-btn').on('click', function(){
 	switchToModellingMode(false);
 	//Cleaning the previous analysis data for new execution
+	global_analysisResult.elementList = "";
```
Akira83: Just added a line to clean the elementList from previous execution.

Issue #47, opened 2017-08-27
Repeating User-Defined Absolute Time Entry.


amgrubb:
The definition of a repeating function states:
> Let a goal g ∈ G, and a sequence of atomic functions A be given, where ai ∈ A is the tuple ⟨goal, type, x, start, stop⟩ and ∀ai ∈ A : ai(goal) = g ∧ ai(stop) = ai+1(start). A repeating compound function is the tuple ⟨g, A, repNum, absTime⟩, where g is the given goal, A is the given set of atomic functions, repNum ∈ N+ is the number of repeats for the entirety of A, and absT ime ∈ N0 is the time between repeats or zero if no value is specified. A repeating compound function is a sequence of atomic functions over disjoint neighbouring intervals that repeat repNum to define how the evaluation of goal g changes over the time.

We need to give the user a way of specifying repNum, absTime.

Perhaps in the function definition here:
![screen shot 2017-08-27 at 11 20 43 am](https://user-images.githubusercontent.com/6596937/29751644-23dbfb90-8b1e-11e7-80bc-e4764b78e1c6.png)
Or perhaps in the absolute values table here:
![screen shot 2017-08-27 at 11 21 33 am](https://user-images.githubusercontent.com/6596937/29751658-536287ee-8b1e-11e7-9ac9-bf256a08cde3.png)

Akira83:
I think that it would be more convenient for the user if this options are set in the Function definition panel, otherwise the user can be lost on where to define this kind of information. But it is just my opinion...

amgrubb:
Yes, but we need to keep all the absolute value together.

amgrubb:
I now agree with @Akira83. The absolute number and number of repeats should be set on the element panel. 

1. We need to add the following change (see screen shot).
![screen shot 2017-10-20 at 9 29 52 pm](https://user-images.githubusercontent.com/6596937/31846883-a9b1f4fe-b5df-11e7-970a-037ef6f4aaad.png)

2. In the Absolute and Relative Assignments table, we need to disable the values that are in the repeating segment. For example, in the Bike Lane Usage example above, we need to disable time points B, C, and D. Because they are in the repeating part.

@g2hchang Can you make these changes?

g2hchang:
Created a pull request #112

g2hchang:
Pull request #112

PR #112, opened 2017-11-06
Repeating UD values

g2hchang:
This is related to #47 
I have implemented users to define number of repeats and absolute length value.
Initially, the repeat counts = 2 and absolute length = 0 as a default value : 
screen shot 2017-11-05 at 11 18 35 pm

Then user can set the value to the value they wish to define as such  : 
screen shot 2017-11-05 at 11 19 23 pm


amgrubb:
Sorry for the delay, are the boxes "text" boxes or are they "up/down" integer boxes?
If "up/down" integer boxes, did you figure out how to show "N/A" in Absolute Length?

g2hchang:
Both boxes are input type="number", so they have "up/down" integer boxes. Also, only way to have "N/A" is to have input type="text". I have added the note to indicate what 0 means in absolute length in red text.

amgrubb:
This is great, I want to add some stuff to the branch before merging.

amgrubb:
@g2hchang  I pushed a few changes. There is still two things I need you to do.
1. InputDynamics.js line 75-76 need to be finished.
2. ElementInspector.js line 143 can you add a bit of whitespace so that the note can naturally break the line after the word Length. "Note: Leave \"Absolute Length\" as 0 for unknown length."

leaf-ui/js/object/InputDynamic.js
```diff
@@ -72,6 +72,10 @@ function getDynamics(){
 				}else{
 					line += "\tR\t" + rBegin + "\t" + rEnd;
 				}
+				// TODO Add the repeat count value and the absolute length value here
+				console.log(typeof $("#repeat-end2").val());
+				line += "\t" + $("#repeat-end2").val() + "\t" + $("#repeat-end3").val();
+				console.log(line);
```
amgrubb: Do you want all these console.log message left in?

Issue #12, opened 2017-08-14
Syntax Checking for Evolving Decomposition


amgrubb:
There are many invalid combinations of decomposition links. For example:
![screen shot 2017-08-14 at 4 48 05 pm](https://user-images.githubusercontent.com/6596937/29291092-6dfcefd0-8110-11e7-96f1-000976559b08.png)

Add syntax checking when you click analysis to show users errors.

Akira83:
Could you clarify this a little bit more? In this example, why is this combination invalid? What I could understand is that in the beginning the Goal_0 can be decomposed in Goal_1 and Goal_2. Then somewhere in next steps it will be only decomposed in Goal_1 without any relationship between Goal_2.

amgrubb:
Will do, this is of lower importance so it might take me a bit of time. I will assign myself so I will get back to it.

g2hchang:
Links with #87

g2hchang:
Pull request #115 merged.

PR #115, opened 2017-11-15
Corrected branch for syntax check Issue #12

g2hchang:
Completed issue #12, which allows user to check if their constraint links are validly assigned to its destination node. For example, if more than 2 nodes pointing to same destination node have different constraint values, then this means it's an invalid link combinations. This feature informs user whether their constraint links combination are valid/not valid.

 **1. User create links.**
screen shot 2017-11-14 at 7 58 13 pm
**2. Select Analysis button, then it triggers to check if links are valid and if not, it highlights all the sets of links that are invalidly combined. Also, it returns error messages and suggestions.**
screen shot 2017-11-14 at 7 52 12 pm

screen shot 2017-11-14 at 7 52 18 pm

**3. If there are more than two groups of links that are invalidly assigned then it will highlight these links as well.**
screen shot 2017-11-14 at 8 02 57 pm

screen shot 2017-11-14 at 8 03 02 pm

**4. If clicked on "Go back to Model View" button, it will return to Model view.**
screen shot 2017-11-14 at 7 52 18 pm

### DIFF
--- a/leaf-ui/css/style.css
+++ b/leaf-ui/css/style.css
@@ -20,7 +20,9 @@ body {
 .header {
    padding: 0px;
 }
-
+#swal2-content b{
+  color: red;
+}
 #name {
   position: absolute;
   top: 0px;
--- a/leaf-ui/index.html
+++ b/leaf-ui/index.html
@@ -135,7 +135,7 @@
 
         <!--SweetAlertJS is used for easy way to customize dialog or alert message
        -->
-        <script src="./plugins/sweetalert/sweetalert.min.js"></script>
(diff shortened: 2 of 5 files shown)


g2hchang:
I need to fix something, please don't merge yet.

g2hchang:
It's not ready to be merged.

amgrubb:
A correction, the contribution links (++,--, -, + etc.) can be merged with decomposition links (and/or). So this is actually valid.

screen shot 2017-11-18 at 4 35 00 pm

I know, it doesn't really make sense.

amgrubb:
Let me know when you have made this correction. :)

Issue #104, opened 2017-10-20
Send "Intermediate Values" to backend


amgrubb:
Dependent on #57 and #93.

Implement SAT-based analysis for intermediate value table.
Figure out how to store and send values to the backend.

Question: How does this work for single path and all paths.

amgrubb:
Created branch feature-userAssignments.

1. Add new class to backend ModelSpec

amgrubb:
@g2hchang I need you to add the values from the table into InputEvaluation.js

User branch feature-userAssignments

lines 9--24, I left you some TODO:

```	//TODO, complete this function.
	//var value =  //get the values from the "Intermediate Values Table"
	//if(values != null){
		//for each value entered into the "Intermediate Values Table" {
			var goal = null;  //TODO
			var absTime = null; //TOOD
			var evaluationValue = null; //TODO
			var evaluation = new InputEvaluation(
					goal,
					absTime,
					evaluationValue);

			evaluations.push(evaluation);
	//	};
	//}

oibrahim22:
We have not looked or changed anything in the backend this summer

leshghi:
SPRINT WEEK OCT 1 @leshghi @epartakki 

**MUST DO:**
- [ ] Create sample model with 4 analysis values to use as an example/test model for Intermediate Values table @epartakki 
- [x] Narrow down all files relevant for Intermediate Values table and implementation in frontend @leshghi 

**SHOULD DO:**
- [ ] Find most recent place in code timeline where Intermediate Values table was working on frontend @leshghi 
- [ ] Compile list of questions for Alicia @epartakki @leshghi 
- [x] Narrow down all files relevant for Intermediate Values table and implementation in backend @epartakki @leshghi 

**STRETCH GOALS:**
- [x] Look for Intermediate Values table bugs
- [ ] Put together plan for how to fix Intermediate values table on frontend

leshghi:
SPRINT WEEK OCTOBER 8 @leshghi @epartakki 
 
- [x] Create sample model with 4 analysis values to use as an example/test model for Intermediate Values table 
- [x] Look for Intermediate Values table bugs
- [x] Put together plan for how to fix Intermediate values table on frontend
- [x] Start working with code to debug frontend Intermediate Values table

leshghi:
SPRINT WEEK OCTOBER 15 @leshghi 

- [x] Fix NT no drop down menu issue

- [x] Reorder numbers at proper timepoints

[...]

PR #116, opened 2017-11-23
Feature user assignments

g2hchang:
This is related issue #104. I am able to generate all the InputEvaluation values.
For example, if the Intermediate Value Tables are as follow : 
screen shot 2017-11-22 at 10 20 01 pm

Then following InputEvaluation lists are generated.
screen shot 2017-11-22 at 10 20 13 pm

When the values are empty, then I am sending to backend as null value i.e.
screen shot 2017-11-22 at 10 23 13 pm
screen shot 2017-11-22 at 10 23 21 pm


amgrubb:
You write the best pull requests! I will look at this hopefully tomorrow or Saturday.

amgrubb:
@g2hchang There is still one issue with the fix.
It currently outputs the following. 
The `evaluationValue` should be 4 bits instead of the english value.
`"evaluationValue": "(FS, T)"` should be `"evaluationValue": "0011"`

```    
"userEvaluations": [
      {
        "goal": "0000",
        "absTime": 2,
        "evaluationValue": "(FS, T)"
      },
      {
        "goal": "0001",
        "absTime": 4,
        "evaluationValue": "(T, FD)"
      }
    ]

PR #119, opened 2017-12-06
UI Bugs

amgrubb:
Added version number and check for null.
Added check if values exist.


g2hchang:
Fixed the bug.
I had to check the case where the user eval table was undefined.
screen shot 2017-12-09 at 12 10 23 pm
screen shot 2017-12-09 at 12 10 04 pm

Issue #97, opened 2017-10-17
Initial satisfaction value in element inspector is cleared when Simulate Single Path is run with > 0 relative time points


fsam95:
Steps to reproduce: 
(1) Create the following model
![init-sat-value-error](https://user-images.githubusercontent.com/8459519/31683422-00db78f0-b34b-11e7-8b52-d7e46cf9afc5.png)

(2) Switch to the analysis tab and set the number of relative time points to 1
(3) Press the Simulate Single Path button
(4) Return to the modeling tab and inspect the state of the Task
**Expected**: initial satisfaction value should be the same (FS)
**Actual**: initial satisfaction value field has been cleared 

It also looks like there's other errors that arise from following this workflow: if the Task element is dragged, both the initial satisfaction value and the function type are set to None. 

Same problem resurfaces when the type of the relationship propagating from an element is changed.

amgrubb:
May have been fixed by #121 
Needs testing. 👍

PR #121, opened 2018-01-02
Bug97 fixed

g2hchang:
Proposed fix for bug 97. Previously, whenever SinglePath is executed, the Initial Satisfaction Values became empty and Function Type values would disappear when you click on a model.
I commented line 819 on release-one branch and it seems to fix it.
screen shot 2018-01-02 at 10 41 44 am


amgrubb:
This was for #97

Issue #59, opened 2017-09-15
Add Buttons to AllPaths popup.


amgrubb:
We need the following buttons:
Accept Current State and Save- takes the current state (and any previous) and adds it to the path generated. (Call single path with values up to the current state.)
Accept Current State and Generate All Next States - stores the current state and then generates the next state.
Close - reject current path and close window.

Akira83:
Button Close current path done.

Akira83:
Accept Current State and Save.
Questions:
In the original analysis (Single Path) I have the following result:
`{
  "elementList": [
    {
      "id": "0000",
      "status": [
        "0000",
        "1110",
        "1110",
        "1110",
        "1100",
        "1110"
      ]
    }
  ],
  "allSolution": [
    
  ],
  "finalAssignedEpoch": [
    "TA0_0",
    "TR1_12",
    "TR3_10",
    "TR2_11",
    "TR5_61",
    "TR4_91"
  ],
  "finalValueTimePoints": [
    "0",
    "10",
    "11",
    "12",
    "61",
    "91"
  ],
  "relativeTimePoints": 5
}`

Then I need to add the result from pop up which is something like this:
`"intentionElements": [
        {
          "id": "0",
          "status": [
            "1110"
          ]
        }
      ]`

My question is what should be the values for finalAssignedEpoch and finalValueTimePoints. I will need to add some values there or the length of each vector is going to be different from the elementsList.

amgrubb:
@Akira83 In the example above in your original simulation, you have only one intention and 5 relative steps, right?

In order to select the next state, we need to ask the user which of the remaining time points they want to apply.

We should meet in person to discuss this.

amgrubb:
- [ ] Create algorithm for resolving time points from All Paths analysis.
Figure out how to determine what the next time point is.
Can we give the user choices.
< Merged with #60 >

amgrubb:
Here is the summary of our meeting today. Let me know if you have any questions.

In the pop up window where the states are displayed, we should also show:
`Path Length`, `Current Step`, `Previous Time`, `Max Time`.

There are three types of time points:
1. Absolute
2. Relative
3. EB Time Points (from intentions and relationship...and may have an absolute value associated).

Once the user clicks `Accept Current State` a small pop up should appear to the user can assign a time point to the current state.

Show all non-assigned EBs in two groups: 1. Those the user can pick, 2. Future time points that cannot be assigned to the current state. 
[...]

PR #134, opened 2018-05-14
#59 Updating View Button

amgrubb:


amgrubb:
@nasirhemed Please review this code. I will merge it when you are done (if it is good).

PR #144, opened 2018-05-15
#142 Syntax check function

nasirhemed:
Change the warning message display for syntax check to give a clearer message 

current

Issue #86, opened 2017-09-29
Test framework integration


g2hchang:
Find suitable testing framework and integrate with current project.
Initial steps would be to include test suite and write some test cases for basic operations.

PR #148, opened 2018-05-17
Issue86testframework

xorye:
This is for #86. Test cases have been reviewed, and appear correct. One test case does not pass, revealing a bug, where adding a single ++ link added first, and then adding a single AND link added second, caused the syntax checker to indicate a syntax error, when there shouldn't be.

This bug has been fixed in the SyntaxFunctionCheck branch, which was for #144.


amgrubb:
@Null7 Please provide a review of this pull request. Do not merge the pull request, just submit a review.

xorye:
I cherry picked from the [SyntaxCheckFunction](https://github.com/amgrubb/BloomingLeaf/tree/SyntaxCheckFunction) branch so this branch has the most recent version of the syntaxCheck() in main.js. I have recently just updated testscript.js so that the test cases reflet the changes made to the syntax checking functions in main.js.

amgrubb:
@Null7 Code was updated, can you confirm your review again. Thanks!

Null7:
Code works fine, but since we are changing the syntax check function further so that it deals with the constraint error better. It is suggested to do the merge request later.

Null7:
Also we are adding cases to check for no relationship, test script needs to be updated.

amgrubb:
@Null7 Let me know when you are done.

Null7:
@amgrubb It's done. Total of 20 test cases tested and passed. Should be good for merging.

Issue #143, opened 2018-05-15
Tool is not responsive sometimes when loaded for the first time


nasirhemed:
Tool does not seem to work sometimes when it is loaded for the first time. The buttons and the nodes do not do anything when they are clicked.

![gif](https://user-images.githubusercontent.com/33851317/40066459-2325cf50-5832-11e8-91f6-e5b5dcedf070.gif)

xorye:
It is difficult to reproduce this error, making it difficult to work on a fix. I will keep my eyes open for the next time this bug happens.

nasirhemed:
Issue fixed

PR #150, opened 2018-05-28
Fix bug where graph fails to load from cookies issue #143

xorye:
… unresponsive

This is for issue #143.

The reason why the webpage was sometimes unresponsive for the first time was that there was an error when the webpage tried to load the graph from the cookies. 

JSON.parse(prevgraph) would throw an error if prevgraph did not contain a proper JSON object in a string format. The cause of the error was simply because a console.log statement called JSON.parse(prevgraph) before checking if prevgraph is a valid JSON in string format.

### DIFF
--- a/leaf-ui/scripts/main.js
+++ b/leaf-ui/scripts/main.js
@@ -142,8 +142,8 @@ if (document.cookie){
 
 	//Loop through the cookies to find the one representing the graph, if it exists
 	for (var i = 0; i < cookies.length; i++){
-		if (cookies[i].indexOf("graph=") >= 0){
-			prevgraph = cookies[i].substr(6);
+		if (cookies[i].indexOf("graph=") !== -1){ // If substring exists
+			prevgraph = cookies[i].substr(cookies[i].indexOf("graph=" + 6)); // Get the substring after graph=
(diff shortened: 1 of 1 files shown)


leaf-ui/scripts/main.js
```diff
@@ -147,10 +147,16 @@ if (document.cookie){
 			break;
 		}
 	}
-	console.log(graph.fromJSON(JSON.parse(prevgraph)));
+
 	if (prevgraph){
-		graph.fromJSON(JSON.parse(prevgraph));
+		try {
```
xorye: Using try/catch as a sanity check.

nasirhemed: Actually, I think it can be fixed in the if statement.
`
if (cookies[i].indexOf("graph=") >= 0){
			prevgraph = cookies[i].substr(6);
`
It is assuming the substring "graph=" is in the beginning of the string. But if it is not then the JSON string is not properly formatted. I propose this instead:

`
if (cookies[i].indexOf("graph=") != 0){
			prevgraph = cookies[i].substr(cookies[i].indexOf("graph=") + 6);
`

PR #153, opened 2018-06-08
Feature user tracking

carolineeeeeee:
To be able to access the database, you need to set up a user-managed webserver on colony and run the code on your server.

Follow the steps here:
1. ssh into colony
2. run **/cs/htdata/user-webserver/makesite**, (check here for reference: https://support.cs.toronto.edu/faq/user_managed_howto.html)
3. cd ~/site/conf, check for the port number in my.cnf
    For example:
     _**[client]
      port               =  40391
      socket           =  /u/boyue/site/conf/run/mysqld.sock**_
     Note: this port is the port number for mysql, the real port number to run the code would be this minus 1 (i.e. the example above should be port 40390)
4. copy leaf-ui into ~/site/www/
scp -r leaf-ui/* boyue@colony.cs.toronto.edu:/u/boyue/site/www/
5. in the browser: www.cs.toronto.edu: \<your port number\> (i.e. the example above should be www.cs.toronto.edu:40390)

Database can be accessed through:
run **mysql --host=$servername --user=$username --password=$password $dbname**
where
$servername = "127.0.0.1:40391";
$username = "root";
$password = "11111111";
$dbname = "my_db";

If you don't need access to the database, you can still run it the normal way in public_html.

### DIFF
--- a/leaf-ui/scripts/ajaxjs.php
+++ b/leaf-ui/scripts/ajaxjs.php
@@ -27,19 +27,20 @@ switch ($type) {
             echo 0;
         }
         break;
-    case "0":
-        $result = $conn->query($query);
-        $result = mysqli_query($conn, $query); //Insert Query
-        if (!$result){
+    case "0": // reading from the database
+        $result_1 = $conn->query($query);
+
+        if (!$result_1){
              echo("Error description: " . mysqli_error($conn));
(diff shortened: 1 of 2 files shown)


amgrubb:
This is the user tacking code.

PR #172, opened 2018-11-23
Fix168

itangdav:
The names of the possible selections for markedValue were different than the possible values for init-sat-value. I just copied and pasted the possible values for init-sat-value into the possible values for markedValue.


amgrubb:
This is good.

Issue #168, opened 2018-10-19
Increase / Decrease Function Issues with None value.


amgrubb:
When adding the Increase or Decrease function (also possibly MP/MN).
If you set it to increase and then leave it as None which is auto-populated, as below:
![screen shot 2018-10-19 at 5 12 48 pm](https://user-images.githubusercontent.com/6596937/47244158-7bdd0a80-d3c2-11e8-8d42-df6eef87775a.png)

then click off the element, and back own, there is no value, as below:
![screen shot 2018-10-19 at 5 13 02 pm](https://user-images.githubusercontent.com/6596937/47244169-84cddc00-d3c2-11e8-9bc1-b7e50a98366d.png)

So the None is not saved.

amgrubb:
This was started in #172, but not complete.

amgrubb:
Fixed in #173.

PR #173, opened 2018-11-27
Update ElementInspector.js

itangdav:
Although I have no idea why it happens, through trial and error, I identified that the markedValue table would not update after the render function is called (basically whenever we click back onto the "cell"). This is even worse as the render actually resets the properties of the cell and then re-updates the cell depending on the input data (which is doubly weird as everything else seems to re-update fine except for the markedValue table). (Note: It would not surprise me if someone else also fails to re-update). Again, I have no idea why, but when I remind the render function to update the markedValue table if the function is either I,D, MP or MN, it seems to fix the problem. But the good thing is that it now works (and I understand why my edit would impact the code and make it work but not why it didn't work in the first place).

### DIFF
--- a/leaf-ui/rappid-extensions/ElementInspector.js
+++ b/leaf-ui/rappid-extensions/ElementInspector.js
@@ -212,6 +212,10 @@ var ElementInspector = Backbone.View.extend({
         } else {
             this.updateHTML(null);
         }
+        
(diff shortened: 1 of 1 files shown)


itangdav:
Hi Alicia, this is the fix to bug 168.

amgrubb:
But #168.

Issue #176, opened 2018-12-20
Deleted Elements remain in the List of Assignments


amgrubb:
I created a big model and ran some simulations.
Then I deleted half the model elements and the deleted elements remained List of Assignments Table (see screenshot of model and View List of Assignments Window).

Branch: Develop

![screen shot 2018-12-20 at 9 27 49 am](https://user-images.githubusercontent.com/6596937/50290560-0cbf7900-043a-11e9-8ceb-22fa2c8f548d.png)

amgrubb:
If you change the value of the evolving function to None before you delete the value, then the issue is resolved. This is an unacceptable workaround.

Connects with deleting element, and the value is still in the model.

amgrubb:
I believe this is fixed now that we got the merge to work.

Issue #166, opened 2018-10-17
Deleted element not properly accounts for in creation of initial model.


amgrubb:
If I start creating a model and delete an element and keep creating a model I get the following error when I click the analysis button:

![screen shot 2018-10-16 at 9 23 59 pm](https://user-images.githubusercontent.com/6596937/47056338-db040a80-d189-11e8-8dec-ba825e36ab81.png)

There error appears to be from the following code.
onFunction.js Lines 38-66:
```
function reassignIntentionIDs() {
	var elements = graph.getElements();
	var intentions = model.intentions;
 	var currID = 0;
	var currIDStr;
	for (var i = 0; i < intentions.length; i++) {
		var intention = intentions[i];
 		if (parseInt(intention.nodeID) !== currID) {
 			// The current intention's ID must be reassigned
		
			// Find the intention's cell
			var cell;
			for (var j = 0; j < elements.length; j++) {
				if (elements[i].attributes.nodeID === intention.nodeID) {
					cell = elements[i];
				}
			}
 			currIDStr = currID.toString();
 			while (currIDStr.length < 4){
	                currIDStr = '0' + currIDStr;
	        }
			cell.attributes.nodeID = currIDStr;
			intention.setNewID(currIDStr);
		}
 		currID += 1;
	}
 	Intention.numOfCreatedInstances = currID;
	Link.numOfCreatedInstances = currID;
}
```

oibrahim22:
Lucy and I tried to recreate the error and we couldn't. We believe that Lucy's solution to issue #176 took care of this one too.

amgrubb:
May be fixed by #180

PR #180, opened 2019-06-06
Fix Issue 176

amgrubb:
#176 -  Lucy fixed this issue.

### DIFF
--- a/leaf-ui/js/object/Class.js
+++ b/leaf-ui/js/object/Class.js
@@ -24,11 +24,11 @@ class Model {
      * @returns {Intention}
      */
     getIntentionByID(nodeID) {
-    	for (var i = 0; i < this.intentions.length; i++) {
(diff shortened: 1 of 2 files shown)


oibrahim22 (approved):
The change in code is more efficient and more importantly functioning. There were  cell.on (remove) functions that were supposed to handle the removing of intentions, actors, and links, they were deleted and moved to the graph.on(remove ) function. This took care of the deleting issues, including issue#166. The problem in that issue(#166) was that elements were not deleted from the list properly, and was causing an undefined error. The only concerning I have about the change is what the previous graph.on(remove) was trying to accomplish. It is clear however, that it was not trying to remove and delete elements.

amgrubb:
I heard you need to make more changes. Let me know when I should review this pull request.

amgrubb:
Oops, I will revert the pull request.

Issue #140, opened 2018-05-14
Allow 'View List of Assignments' to be view from Modeling View.


amgrubb:
Can we move the button because this is stored with the model.

oibrahim22:
There is a current pull request to add this change

oibrahim22:
Will not fix. This can be updated if we redesign the representation of the model and analysis. Not as simple as adding a button. It requires a transformation/refactoring of the analysisInspector.

PR #191, opened 2019-07-19
Allow 'View List of Assignments' to be view from Modeling View. #140

oibrahim22:
These changes move view list of assignment list button from analysis view to modeling view by changing the code that controls it from AnalysisInspector.js to ElementInspector.js


oibrahim22:
Starting over

PR #192, opened 2019-07-25
issue 94

oibrahim22:
change to implement issue 94. Right now if there is a cycle detected, the nodes in the cycle change color. If one node is in more than one cycle its color changes occurring to the first cycle detected


amgrubb (approved):
Nice work!

Issue #197, opened 2020-02-05
Clear Full Model button only clearing some


kmbspencer:
Clear full model button had to be pressed 2-3 times to clear the full model when I created a new model from scratch. On kate-branch (branched from develop).

kmbspencer:
fixed with newest pull request

PR #199, opened 2020-06-08
clear full model now clears, but still leaves traces of elements

amgrubb:
clear full model will make all elemnts dissapear, but when the json is saved and read there are still the intention names


amgrubb:
Fixes issues #197

mvarnum (approved):
Looks good.

Issue #203, opened 2020-06-12
Deleting nodes yield an "Uncaught TypeError"


mvarnum:

kmbspencer:
fixed with newest pull

PR #204, opened 2020-06-12
Deleting nodes no longer yields error -- node's dynamic function and links are successfully removed

mvarnum:


mvarnum:
Fixes issue #203

kmbspencer (approved):
looks great! deleted many nodes and never got any errors

PR #206, opened 2020-06-16
fixed conflict with deleteBug

kmbspencer:
issues #203 and #199


mvarnum (approved):
Tested on a few models, looks good.

Issue #208, opened 2020-06-18
Add error messages for Solve Next State


amgrubb:
Cannot be called from last state and Solve Single Path MUST be called first.

PR #217, opened 2020-08-03
Added error messages when user selects "Explore Next States" before simulating a single path or with the last time point selected

mvarnum:


mvarnum:
Fixes #208

kmbspencer (approved):

PR #224, opened 2020-11-05
temp directory error fix

leshghi:
Automatically creates temp directory if the temp folder has not been manually created. Will fix error caused on first download of develop branch where temp directory has not been created yet and program crashes until user manually creates it.


esterzhao (approved):
Looks good, thanks for the fix

Issue #238, opened 2020-11-30
Update Assignments Table Unassign button to remove added time points from analysis if applicable


leshghi:
When you add specific time points to UD functions in the Assignment Table, they appear on the IVT. This is good and makes sense. However, when you click the “Unassign” on the Assignments Table, the time point will remain in the IVT table even if it is not being used anywhere else. This means the IVT very quickly gets clogged up with time points that are not necessary. This fix should ideally refresh the IVT when values are unassigned in the Assignment Table so that the IVT does not get so busy with unnecessary values.

PR #260, opened 2021-01-15
Fixes leftover time points in IVT

leshghi:
By moving the analysisRequest.absTimePts update to the assignment save table, it correctly updates when a time point has been removed from the absolute intentions and considers that removal while updating analysisRequest.absTimePoints. This fixes the unused/removed absolute intention time points crowding up the IVT as per issue #238

Also makes a small change with the absolute time points field so that it saves to the model object after “save” is clicked on the assignments table instead of immediately after updating. This is just so it is consistent with how the rest of the table works. Since there is a save button, stuff should only get saved when it is pressed.

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -24,21 +24,6 @@ $('#max-abs-time').on('change', function(){
     }
 });
 
-/** 
- * Sets Absolute time points
- */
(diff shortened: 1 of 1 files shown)


esterzhao (approved):
Looks good to me @leshghi! I did a quick test run to ensure that its working, and I couldn't find any issues with the run.

Issue #263, opened 2021-01-27
Reset to a blank model UI when switching from result to config


leshghi:
When you first open the analysis page on a model you see a clean model with no results:
<img width="1176" alt="Screen Shot 2021-01-27 at 3 10 07 PM" src="https://user-images.githubusercontent.com/43756305/106047650-dc834380-60b1-11eb-82d2-0feec91e92be.png">
After running a simulation, results appear on the model with a slider:
<img width="1171" alt="Screen Shot 2021-01-27 at 3 11 42 PM" src="https://user-images.githubusercontent.com/43756305/106047724-f886e500-60b1-11eb-9a64-18e09ef6a308.png">
But when you switch back to a different configuration, the results are still populated on the UI:
<img width="1174" alt="Screen Shot 2021-01-27 at 3 10 23 PM" src="https://user-images.githubusercontent.com/43756305/106047774-076d9780-60b2-11eb-84f0-297b6e6dc3eb.png">

Need code to reset to the blank model UI when switching from a result to a configuration to avoid user confusion about which result is associated with which config. This does not cause any confusion in the backend or with the map, but it does impact the user experience and should be fixed asap.

krhablutzel:
Just pushed updates on branch `mm-reset-ui` to hide slider, rest analysis values, and change EVO slider when switching to a configuration

Issue #262, opened 2021-01-27
Results bar is longer than configuration bar in analysis left sidebar


leshghi:
See attached picture.
Results and config should be equal length.
<img width="204" alt="Screen Shot 2021-01-27 at 3 04 36 PM" src="https://user-images.githubusercontent.com/43756305/106047043-199b0600-60b1-11eb-8f36-0b5b3e399ce6.png">

Issue #261, opened 2021-01-27
Have IVT and Assignments Table populate with correct configuration data


leshghi:
Currently the IVT and Assignments Table is not properly reflecting the current configuration selection when you switch between AnalysisConfigurations in the left hand sidebar in analysis mode. A function is needed to update the table and modal box when a new configuration is created or clicked.

PR #264, opened 2021-01-27
Analysis config sidebar

### LINKED ISSUE #250: Create analysis configurations/results sidebar
* Update the front-end analysis view to include a sidebar where analysis configurations and results can be viewed and saved
* Make a decision on how to save analysis configurations -> autosave when switching between them, or required save button? How do you save the info on each configuration? Local hosted continuously updated list? **Alicia's note: analysis configurations are only saved once analysis is actually run**
* Generally update the UI to account for model management changes (ex. Scrape out history log, see if anything is usable)

leshghi:
This is the first draft of a rough implementation/outline of the UI for the analysis configuration sidebar. It still contains many outstanding bugs including but not nearly limited to #261, #262, and #263. Future iterations will add load and save functionality, as well deletion, renaming, and warnings when the model is changed. However, for a first draft this should be sufficient for the rest of the project to work off of to add more functionality and fix bugs.

### DIFF
--- a/leaf-ui/js/analysis.js
+++ b/leaf-ui/js/analysis.js
@@ -15,12 +15,12 @@ var analysisResult;
 var savedAnalysisData;
 var graph;
 var current;
-// Analysis Configuration map (key: configId, value: analysisConfig object)
-var analysisMap = new Map();
-// Global variable to keep track of what analysis configuration is currently being used
-var currAnalysisConfig;
-// Count for number of analysis configurations
-var configCount = 0;
+// // Analysis Configuration map (key: configId, value: analysisConfig object)
+// var analysisMap = new Map();
+// // Global variable to keep track of what analysis configuration is currently being used
+// var currAnalysisConfig;
+// // Count for number of analysis configurations
+// var configCount = 0;
 
 var model;
 
--- a/leaf-ui/js/analysis.js
+++ b/leaf-ui/js/analysis.js
@@ -15,12 +15,8 @@ var analysisResult;
 var savedAnalysisData;
 var graph;
 var current;
-// // Analysis Configuration map (key: configId, value: analysisConfig object)
-// var analysisMap = new Map();
-// // Global variable to keep track of what analysis configuration is currently being used
-// var currAnalysisConfig;
-// // Count for number of analysis configurations
-// var configCount = 0;
+// TODO: AnalysisConfig variables moved to displayAnalysis
+// Need to come back to determine if location of othe vars is correct
 
 var model;
 
--- a/leaf-ui/js/backendComm.js
+++ b/leaf-ui/js/backendComm.js
@@ -102,10 +102,10 @@ function responseFunc(isGetNextSteps, response){
 				analysisResult = convertToAnalysisResult(results);
 				displayAnalysis(results);
 
-				// Update results in analysis sidebar
-				updateResults();
 				 // Save result to the corresponding analysis configuration object
 				 currAnalysisConfig.addResult(convertToAnalysisResult(results));
+				 // Update results in analysis sidebar
+				updateResults();
 				 // Add the analysisConfiguration to the analysisMap for access in the analysis config sidebar
 				 analysisMap.set(currAnalysisConfig.id, currAnalysisConfig);
 			 }
--- a/leaf-ui/js/backendComm.js
+++ b/leaf-ui/js/backendComm.js
@@ -97,11 +97,13 @@ function responseFunc(isGetNextSteps, response){
 					console.log("in backendcomm, saving all next state results");
 					 open_analysis_viewer();
 			} else {
-				var resultsString = JSON.stringify(results);
-				 console.log(JSON.stringify(results)); 
-				 savedAnalysisData.singlePathResult = results;
+				console.log(JSON.stringify(results)); 
+				savedAnalysisData.singlePathResult = results;
 				analysisResult = convertToAnalysisResult(results);
 				displayAnalysis(results);
+
+				// Update results in analysis sidebar
+				updateResults();
 				 // Save result to the corresponding analysis configuration object
 				 currAnalysisConfig.addResult(convertToAnalysisResult(results));
 				 // Add the analysisConfiguration to the analysisMap for access in the analysis config sidebar
@@ -172,7 +174,7 @@ function getFileResults(isGetNextSteps){
                     analysisResult.allSolution = results.allSolution;
                     analysisRequest.previousAnalysis = analysisResult;
                     console.log("previousAnalysis");
-                    console.log(analysisRequest.previousAnalysis);
+					console.log(analysisRequest.previousAnalysis);
 					displayAnalysis(results);
 				}
 			}
--- a/leaf-ui/js/backendComm.js
+++ b/leaf-ui/js/backendComm.js
@@ -103,7 +103,7 @@ function responseFunc(isGetNextSteps, response){
 				displayAnalysis(results);
 
 				 // Save result to the corresponding analysis configuration object
-				 currAnalysisConfig.addResult(convertToAnalysisResult(results));
+				 currAnalysisConfig.addResult(analysisResult);
 				 // Update results in analysis sidebar
 				updateResults();
 				 // Add the analysisConfiguration to the analysisMap for access in the analysis config sidebar
--- a/leaf-ui/js/backendComm.js
+++ b/leaf-ui/js/backendComm.js
@@ -100,29 +100,12 @@ function responseFunc(isGetNextSteps, response){
 				var resultsString = JSON.stringify(results);
 				 console.log(JSON.stringify(results)); 
 				 savedAnalysisData.singlePathResult = results;
-				//  analysisResult.assignedEpoch = results.assignedEpoch;
-				//  analysisResult.timePointPath = results.timePointPath;
-				//  analysisResult.timePointPathSize = results.timePointPathSize;
-				//  analysisResult.elementList = results.elementList;
-				//  analysisResult.allSolution = results.allSolution;
-				//  analysisRequest.previousAnalysis = analysisResult;
-				//  console.log("previousAnalysis");
-				//  console.log(analysisRequest.previousAnalysis);
-				//  displayAnalysis(results);
-
-				//  analysisResult.colorVis = new EVO(results.elementList);
-				//  analysisResult.isPathSim = true;
-				//  analysisResult.colorVis.singlePathResponse(results.elementList);
 				analysisResult = convertToAnalysisResult(results);
 				displayAnalysis(results);
-				// TODO: Need to make a new object to add to the results list to avoid overwriting the same global variable each time
-				 // Save analysisResult to the corresponding analysis configuration object
+				 // Save result to the corresponding analysis configuration object
 				 currAnalysisConfig.addResult(convertToAnalysisResult(results));
 				 // Add the analysisConfiguration to the analysisMap for access in the analysis config sidebar
 				 analysisMap.set(currAnalysisConfig.id, currAnalysisConfig);
-				 //TODO remove this
-				 //confirm how many results there are 
-				 console.log("Num analysis results: " + analysisMap.get(currAnalysisConfig.id).analysisResults.length);
 			 }
 		 }
 	 }
--- a/leaf-ui/js/backendComm.js
+++ b/leaf-ui/js/backendComm.js
@@ -100,33 +100,33 @@ function responseFunc(isGetNextSteps, response){
 				var resultsString = JSON.stringify(results);
 				 console.log(JSON.stringify(results)); 
 				 savedAnalysisData.singlePathResult = results;
-				 analysisResult.assignedEpoch = results.assignedEpoch;
-				 analysisResult.timePointPath = results.timePointPath;
-				 analysisResult.timePointPathSize = results.timePointPathSize;
-				 analysisResult.elementList = results.elementList;
-				 analysisResult.allSolution = results.allSolution;
-				 analysisRequest.previousAnalysis = analysisResult;
-				 console.log("previousAnalysis");
-				 console.log(analysisRequest.previousAnalysis);
-				 displayAnalysis(results);
+				//  analysisResult.assignedEpoch = results.assignedEpoch;
+				//  analysisResult.timePointPath = results.timePointPath;
+				//  analysisResult.timePointPathSize = results.timePointPathSize;
+				//  analysisResult.elementList = results.elementList;
+				//  analysisResult.allSolution = results.allSolution;
+				//  analysisRequest.previousAnalysis = analysisResult;
+				//  console.log("previousAnalysis");
+				//  console.log(analysisRequest.previousAnalysis);
+				//  displayAnalysis(results);
 
-				 analysisResult.colorVis = new EVO(results.elementList);
-				 analysisResult.isPathSim = true;
-				 analysisResult.colorVis.singlePathResponse(results.elementList);
-
+				//  analysisResult.colorVis = new EVO(results.elementList);
+				//  analysisResult.isPathSim = true;
+				//  analysisResult.colorVis.singlePathResponse(results.elementList);
+				analysisResult = convertToAnalysisResult(results);
+				displayAnalysis(results);
+				// TODO: Need to make a new object to add to the results list to avoid overwriting the same global variable each time
 				 // Save analysisResult to the corresponding analysis configuration object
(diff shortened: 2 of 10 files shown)


leaf-ui/js/analysis.js
```diff
@@ -15,12 +15,12 @@ var analysisResult;
 var savedAnalysisData;
 var graph;
 var current;
-// Analysis Configuration map (key: configId, value: analysisConfig object)
-var analysisMap = new Map();
-// Global variable to keep track of what analysis configuration is currently being used
-var currAnalysisConfig;
-// Count for number of analysis configurations
-var configCount = 0;
+// // Analysis Configuration map (key: configId, value: analysisConfig object)
```
esterzhao: Since you directly copy pasted this into displayAnalysis, it might be better for now to delete these commented out lines and instead leave a TODO note saying that the analysis configuration variables moved to displayAnalysis, since we need to come back to this to figure out the importance of the other vars that live here and how they integrate into what we're doing

leaf-ui/js/backendComm.js
```diff
@@ -353,3 +338,17 @@ function getIDs(backendErrorMsg) {
 
 	return arr;
 }
+
+function convertToAnalysisResult(results){
```
esterzhao, line 342: @amgrubb Do you know if it would disrupt the setup if we added this as a class method to the `AnalysisResult` class in `leaf-ui/js/object/Class.js`? Just wondering whether we can streamline a little, but I'm not sure if that would affect communication with the backend at all. Still very shaky on gson

leaf-ui/js/backendComm.js
```diff
@@ -97,37 +97,22 @@ function responseFunc(isGetNextSteps, response){
...
-
-				 // Save analysisResult to the corresponding analysis configuration object
-				 currAnalysisConfig.addAnalysis(analysisResult);
+				console.log(JSON.stringify(results)); 
+				savedAnalysisData.singlePathResult = results;
+				analysisResult = convertToAnalysisResult(results);
+				displayAnalysis(results);
+
+				 // Save result to the corresponding analysis configuration object
+				 currAnalysisConfig.addResult(convertToAnalysisResult(results));
```
esterzhao: any reason for not simplifying? no need to call `convertToAnalysisResult()` twice
```suggestion
				 currAnalysisConfig.addResult(analysisResult);
```

leaf-ui/js/backendComm.js
```diff
@@ -97,37 +97,22 @@ function responseFunc(isGetNextSteps, response){
...
+				 // Update results in analysis sidebar
+				updateResults();
 				 // Add the analysisConfiguration to the analysisMap for access in the analysis config sidebar
-				 analysisMap[currAnalysisConfig.id] = currAnalysisConfig;
-				 console.log("Saved analysis configuration and result");
-				 //TODO remove this
-				 //confirm how many results there are 
-				 console.log("Num analysis results: " + analysisMap[currAnalysisConfig.id].analysisResults.length);
+				 analysisMap.set(currAnalysisConfig.id, currAnalysisConfig);
```
esterzhao, line 110: thanks for the fix!

leaf-ui/js/backendComm.js
```diff
@@ -189,7 +174,7 @@ function getFileResults(isGetNextSteps){
                     analysisResult.allSolution = results.allSolution;
                     analysisRequest.previousAnalysis = analysisResult;
                     console.log("previousAnalysis");
-                    console.log(analysisRequest.previousAnalysis);
+					console.log(analysisRequest.previousAnalysis);
```
esterzhao, line 177: stop being dramatic github

leaf-ui/js/displayAnalysis.js
```diff
@@ -3,6 +3,15 @@
  * that the web application would receive from the back end.
```
esterzhao, line 3: Probably want to update this docstring to include reference to the analysis config bar eventually

leaf-ui/js/displayAnalysis.js
```diff
@@ -31,7 +40,11 @@ function displayAnalysis(analysisResults){
     // ElementList = analysisResults.elementList;
 
     // Update history log
-    updateHistory(currentAnalysis);
+    // updateHistory(currentAnalysis);
```
esterzhao, line 42: @mvarnum did we make a decision on whether we are completely scrapping the history log? I can't remember -- if we basically are we should remove this line of code to clean up

leaf-ui/js/displayAnalysis.js
```diff
@@ -31,7 +40,11 @@ function displayAnalysis(analysisResults){
...
+
+    // Update Analysis Config bar
```
esterzhao: What is meant to go in this empty space? Or was this comment supposed to be deleted?

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
 
     var log = new logObject(currentAnalysis, 0);
     historyObject.allHistory.push(log);
-}
\ No newline at end of file
+}
+
+/**
+ * Function to set up the initial config
```
esterzhao:
```suggestion
 * Function to set up the initial analysis configuration upon page load
```

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
...
+    $("#analysis-sidebar").append(
+        '<button class="log-elements" id="'+currAnalysisConfig.id+'" style="padding: 12px; font-size: 16px; border: none; outline: none; background-color:#A9A9A9">' + currAnalysisConfig.id + '</button><div style="position:absolute; display:inline-block"><button class="dropdown" id= "'+buttonLabel+'" style="padding: 12px; font-size: 16px; height: 42px; border: none; outline: none;"><i class="fa fa-caret-down fa-2x" style="cursor:pointer;"></i></button>'
+        + '</div><div class = "dropdown-container" id="'+label+'"></div>'
+      );
+
+    console.log(analysisMap)
+}
+
+/**
+ * Adds a new analysisConfig to the sidebar
```
esterzhao:
```suggestion
 * Adds a new analysis configuration to the Config sidebar
```

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
...
+function addAnalysisConfig(){
+    $(".log-elements").css("background-color", "");
+    $(".result-elements").css("background-color", "");
+    // Update current config with current analysisRequest and set the udpated config in map
+    currAnalysisConfig.updateAnalysis(analysisRequest);
+    analysisMap.set(currAnalysisConfig.id, currAnalysisConfig);
+
+    // Figure out number of new config, name and create it, and then add it to the map
+    var numElements = analysisMap.size;
+    var id = "Configuration" + (numElements+1).toString()
```
esterzhao: I don't think we need the `numElements` var
```suggestion
    var id = "Configuration" + (analysisMap.size+1).toString()
```

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
...
+
+    var newConfig = new AnalysisConfiguration(id, newRequest);
+    analysisMap.set(id, newConfig);
+
+    // Update current config to be the new config, and update analysisRequest to match new config
+    currAnalysisConfig = newConfig;
+    analysisRequest = currAnalysisConfig.getAnalysisRequest();
+
+    // Reset analysis sidebar to default
+    updateAnalysisBar();
```
esterzhao: i would rename this func `refreshAnalysisBar()`

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
...
+ */
+function updateResults(){
+    $(".result-elements").css("background-color", "");
+    var label = currAnalysisConfig.id + "-dropdown";
+    var resultCount = analysisMap.get(currAnalysisConfig.id).analysisResults.length;
+    var id = "Result " + resultCount;
+    document.getElementById(label).insertAdjacentHTML("beforeend","<a class='result-elements' style='background-color:#A9A9A9''>" + id + "</a>");
+}
+
+function updateAnalysisBar(){
```
esterzhao: see above comment - rename? optional

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
...
+    $('#num-rel-time').val(analysisRequest.numRelTime);
+    document.getElementById('conflict-level').value = analysisRequest.conflictLevel;
+}
+
+/**
+ * Switches between analysis configurations
+ */
+$('#analysis-sidebar').on("click", ".log-elements", function(e){
+    //Save and/or update past analysis config in map
+    currAnalysisConfig.updateAnalysis(analysisRequest)
```
esterzhao, line 335: smart move to update when you're switching!

leaf-ui/js/displayAnalysis.js
```diff
@@ -239,4 +252,147 @@ function updateHistory(currentAnalysis){
...
+    analysisMap.set(currAnalysisConfig.id, currAnalysisConfig);
+    currentConfigHtml = e.target;
```
esterzhao: this isn't a critique I just want you to maybe explain to me how this works because html really confuses me lol - I saw you initialized this global var up above, what exactly is it for? is it like the button location or something?

leaf-ui/js/objects.js
```diff
@@ -207,6 +207,7 @@ class AnalysisConfiguration {
 
 	constructor(id, analysisRequest) {
 		this.id = id;
+		this.analysisRequest = analysisRequest;
```
esterzhao, line 210: thank you for updating this, its v clean! love it

leaf-ui/js/objects.js
```diff
@@ -221,7 +222,30 @@ class AnalysisConfiguration {
...
 	 */
-	addAnalysis(analysisResult) {
+	addResult(analysisResult) {
 		this.analysisResults.push(analysisResult);
 	}
+
+	/**
+	 * Updates Config Values from AnalysisRequest
+	 */
+	updateAnalysis(analysisRequest){
```
esterzhao, line 232: I feel like the word "analysis" in this class implies the `AnalysisResult`, but I'm not sure what exactly is the best method name. For now I'll just suggest "update"
```suggestion
	update(analysisRequest){
```

leaf-ui/js/onFunctions.js
```diff
@@ -1343,3 +1348,9 @@ function checkForMultipleNB(node) {
 	return num >= 1;
 }
 
+function loadAnalysisConfig(){
```
esterzhao, line 1356: Is this func exclusively for loading the first config when you switch to analysis mode, or do you imagine that it could be used for loading additional configs from a json? Just clarifying

esterzhao: Also good practice to docstring all functions

leaf-ui/rappid-extensions/AnalysisInspector.js
```diff
@@ -96,12 +96,12 @@ var AnalysisInspector = Backbone.View.extend({
 		analysisRequest.action = "singlePath";
 		analysisRequest.currentState = "0|0";
 
-		// TODO: add a check to see if this configuration already exists
-		// Save the current analysis configuration
-		currAnalysisConfig = new AnalysisConfiguration(configCount, analysisRequest);
-		// Increase the number of configurations
-		configCount += 1;
-		console.log("Num configurations: " + configCount);
+		// // TODO: add a check to see if this configuration already exists
```
esterzhao: You can delete this

esterzhao (approved):
Did an initial run through of the code, haven't tested it in browser yet, but if you're ok with that then consider this an approval -- I figure since we're just trying to get this merged to model-management ASAP so more people can bug fix this is good enough. My comments are almost all naming conventions/requests for comments/clarifying qs

leaf-ui/js/analysis.js
leshghi: Done

leaf-ui/js/backendComm.js
leshghi: Definitely don't think this change would impact anything with gson/json. However, bit confused on purpose of making it a class method in the AnalysisResult object. Currently this code is creating a new AnalysisRequest object from a parsed json file. Its necessary to create a new object to avoid the issue we ran into with last pull request re: overwriting global variables in the config lists. Would you rather we make a new blank AnalysisRequest and then call this method as a class method (maybe renamed to something like convertResults) to populate the values in the new object? I think that could definitely be done, just not clear if that is what you are suggesting.

leaf-ui/js/backendComm.js
leshghi: No reason at all, thanks for the suggestion!

leaf-ui/js/backendComm.js
leshghi: Yeah lmfao I went back when I saw this to be like what is going on and the extra spaces arent even showing up in my IDE lol

leaf-ui/js/displayAnalysis.js
leshghi: Done, will add more detail in future iterations once code is more set

leaf-ui/js/displayAnalysis.js
leshghi: Supposed to be deleted, thanks for the catch!

leaf-ui/js/displayAnalysis.js
leshghi: Done

leaf-ui/js/displayAnalysis.js
leshghi: Done

leaf-ui/js/displayAnalysis.js
leshghi: Done thank you!

leaf-ui/js/displayAnalysis.js
leshghi: Done

leaf-ui/js/displayAnalysis.js
leshghi: Done

leaf-ui/js/displayAnalysis.js
leshghi: Yeah it basically tracks like what html element is being clicked currently based on the event e. But I actually think its not necessary, I ended up streamlining the html stuff a tiny bit so can delete. Thanks for pointing out!

leaf-ui/js/objects.js
leshghi: Yeah I figured since we ended up needing to grab the analysisRequest from the analysisConfig so often it might be easier to just save directly. Left in the other values for the Config class for now though. I think its mostly a design tradeoff between the efficiency of being able to save and get the whole request faster vs not storing more information from the request than necessary. AnalysisRequest currently contains the previous step value which we need for next states so once we get there we can probably discuss in more detail what parts of the Config class we need and what we don't.

leaf-ui/js/objects.js
leshghi: I feel like update is too vague (could suggest id as well which could cause confusion once we implement naming, doesn't say what you are updating explicitly, etc) but hear your concern. Thoughts on updateAnalysisValues ?

leaf-ui/js/onFunctions.js
leshghi: For now its exclusively for loading the first config but my hope is in the future once we do loading and saving it can be used to check for additional configs and load all of them. Docstring added!

leaf-ui/rappid-extensions/AnalysisInspector.js
leshghi: Done! Thanks for the catch totally forgot this was ever in there oops

leaf-ui/js/objects.js
esterzhao: how about updateAnalysisConfig?

esterzhao (approved):

leaf-ui/js/backendComm.js
esterzhao: Gotcha - I misunderstood the purpose, I thought it was just wiping the AnalysisRequest object params and repopulating it with new vars, in which case a class func would make more sense. This is fine!

PR #271, opened 2021-02-04
bug fix - Show loaded results on click 

esterzhao:
Two bugs one stone: allow users to view results in the UI by clicking on them, and fix a new bug I found - the analysis sidebar needs to be refreshed when loading data from a json, so as to display the info in the AnalysisRequest instead of the default conflict prevention level etc. when switching to analysis mode for the first time.


krhablutzel:
This all looks great!

leshghi (approved):
Looks good to me! Nice job :)

PR #276, opened 2021-02-09
Issue 272

### LINKED ISSUE #261: Have IVT and Assignments Table populate with correct configuration data

leshghi:
Update so that the IVT and the analysis values in the Assignments modal correctly reflect the current analysis config. Also fixed an outstanding bug in the IVT where it would not re-render the saved values after being saved/closed and instead reset the UI to a blank option. This should close #261.

### DIFF
--- a/leaf-ui/js/displayAnalysis.js
+++ b/leaf-ui/js/displayAnalysis.js
@@ -289,8 +289,6 @@ function loadAnalysis(){
         updateResults();
         console.log("analysis request:")
         console.log(analysisRequest);
-        // Refresh the sidebar to include the config vars
-        refreshAnalysis();
     }
     // TODO: figure out how to set it to the element of the map that will populate on top
     currAnalysisConfig = analysisMap.get("Configuration1");
@@ -299,6 +297,7 @@ function loadAnalysis(){
     analysisRequest = currAnalysisConfig.analysisRequest;
     console.log("analysis request:")
     console.log(analysisRequest);
+    // Refresh the sidebar to include the config vars
     refreshAnalysisUI();
 }
 
@@ -316,7 +315,7 @@ function addNewAnalysisConfig(){
     // default Analysis Request needed for now for user assignments list
     // TODO: Look into perserving base UAL throughout analysisRequests
     var newRequest = new AnalysisRequest();
-    newRequest.userAssignmentsList = defaultUAL;
+    defaultUAL.forEach(userEval => newRequest.userAssignmentsList.push(userEval));
 
     var newConfig = new AnalysisConfiguration(id, newRequest);
     analysisMap.set(id, newConfig);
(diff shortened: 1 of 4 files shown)


leaf-ui/js/displayAnalysis.js
```diff
@@ -277,10 +277,13 @@ function loadAnalysis(){
...
         console.log(analysisRequest);
     }
+    // TODO: figure out how to set it to the element of the map that will populate on top
+    currAnalysisConfig = analysisMap.get("Configuration1");
     // Set default UAL to preserve in future configs
     defaultUAL = currAnalysisConfig.userAssignmentsList;
+    analysisRequest = currAnalysisConfig.analysisRequest;
     // Refresh the sidebar to include the config vars
-    refreshAnalysisBar();
+    refreshAnalysisUI();
```
esterzhao, line 286: much needed edit, thanks for this *in the distance, my branch cries in merge conflict*

leaf-ui/js/displayAnalysis.js
```diff
@@ -297,7 +300,7 @@ function addNewAnalysisConfig(){
     // default Analysis Request needed for now for user assignments list
     // TODO: Look into perserving base UAL throughout analysisRequests
```
esterzhao, line 301: looks like you fulfilled this TODO if I'm following correctly, in which case remove this comment

leaf-ui/js/objects.js
```diff
@@ -239,13 +239,14 @@ class AnalysisConfiguration {
 	 * Updates Config Values from AnalysisRequest
 	 */
 	updateAnalysis(analysisRequest){
+		this.analysisRequest = analysisRequest;
```
esterzhao, line 242: git is such a spaz 🙄

leaf-ui/rappid-extensions/AnalysisInspector.js
```diff
@@ -430,18 +430,23 @@ var AnalysisInspector = Backbone.View.extend({
...
-					selectElement.val(intEval.evaluationValue);
-				}
 				selectElement.append(options);
 				selectTd.append(selectElement);
 				row.append(selectTd);
-
 				}
 			$('#interm-list').append(row);
 		}
+		// Needed after looping through and appending rows to avoid voiding the select
```
esterzhao, line 439: can't say I understand this... def due to my unfamiliarity with this code but can you clarify (also in comments)

esterzhao (approved):
Looks good to me, no new unknown bugs found while testing. Would like to have @ibrodycalixto or someone else take a look before merging to confirm the IVT/Assignments modal works as intended since I am pretty shaky on those

leaf-ui/rappid-extensions/AnalysisInspector.js
leshghi: Yes! Its just a weird way javascript working with html is set up which basically means even though before it was correctly setting the selectElement.val, that value wasn't propagating when adding selectElement into selectTd and then into the actual row. Instead we needed to build the row, and Then set the selected value inside that row using "this" to access the corresponding select menu.

leaf-ui/js/objects.js
leshghi: trulyyyy

krhablutzel (commented):
This looks great to me as well! I appreciate the function name change with `refreshAnalysisUI()`. Doesn't appear to add any new bugs

PR #277, opened 2021-02-10
Load new file bug

### LINKED ISSUE #266: Allow new model+analysis to be loaded without refreshing the page
If you do not fully refresh the page: When loading a new model+analysis json file from the Model view, the Analysis view Configurations bar is not cleared before repopulating, so there end up being duplicates of the Analysis Configurations in the UI.

If you refresh, there are no issues.

Screen Shot 2021-02-01 at 1 10 34 PM

esterzhao:
Bug fix for issue #266 -- created some additional html containers to separate different config objects so that they can be cleared out when a new file is loaded, and that so eventually we can implement a config delete button.

### DIFF
--- a/leaf-ui/js/displayAnalysis.js
+++ b/leaf-ui/js/displayAnalysis.js
@@ -8,8 +8,6 @@
 var analysisMap = new Map();
 // Global variable to keep track of what analysis configuration is currently being used
 var currAnalysisConfig;
-// Count for number of analysis configurations
-var configCount = 0;
 
 /**
  * Displays the analysis to the web app, by displaying the slider and the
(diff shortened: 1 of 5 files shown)


leaf-ui/js/displayAnalysis.js
```diff
@@ -320,12 +313,36 @@ function addAnalysisConfig() {
...
         + '</div><div class = "dropdown-container" id="'+label+'"></div>'
+        + '</div>'
       );
 }
 
+/**
+ * Removes an analysis configuration from the analysisMap and UI sidebar
+ * TODO implement with a button
+ */
+function removeConfiguration() {
```
leshghi, line 332: Do you ever call this function or is it going to be used later when we implement delete? If it is going to be used for delete we should probably have it be passed the id of the config to delete instead of having it always delete the current one and/or tie it directly to a click event and access the clicked element from there.

leshghi (approved):
This looks great! I think in future iterations we may want to see if there is another way to do this that doesn't involve making additional containers (just to simplify html), but this is also a really smart concise solution so maybe not! I found one bug while testing but I think its outstanding and not caused by this branch at all so will just add to our issues backlog.  Awesome job :)

leaf-ui/js/displayAnalysis.js
krhablutzel: If I'm not mistaken, this function was part of the initial approach to clearing the analysis config sidebar for loading a new model, but isn't needed with the current fix. I do think we'll need a removeConfiguration() eventually, but I agree that it could be useful to be able to pass a config id

krhablutzel (approved):
Looks great! All the nested html tags appear to be ordered correctly, and I appreciate the ability to clear all configurations in one line. I also appreciate having a separate function for adding the first analysisConfig into the UI; thanks for building this!

leaf-ui/js/displayAnalysis.js
```diff
@@ -320,12 +313,36 @@ function addAnalysisConfig() {
...
+    var configDiv = document.getElementById(currAnalysisConfig.id);
+    configDiv.remove();
+    // Remove config from analysisMap
+    analysisMap.delete(currAnalysisConfig.id);
+}
+
+/**
+ * Clears the analysis config sidebar
+ */
+function clearAnalysis() {
```
krhablutzel: This function name might be a little vague, with all the other types of analysis. Maybe `clearAnalysisConfigSidebar()`? Wordy, I know

leaf-ui/js/displayAnalysis.js
esterzhao: Yes so I was going to use this func before and ended up not needing it, but since we are planning on a delete functionality I left it there. I will update it to pass a config id in once I get the feature implemented! I was previously thinking that on a click event to delete you should be looking at the config you want to delete, so the currConfig will be set to the one you want to delete. Didn't fully implement the feature on this branch because I didn't want to tangle too much up into this pull request.

leaf-ui/js/displayAnalysis.js
esterzhao: changed! :)

Issue #282, opened 2021-02-16
Last config added incorrectly shows as selected for loaded models


leshghi:
When you load in a model with analysis configs, the analysis view will show the last configs loaded as selected. However, the actual currAnalysisConfig will be the first configuration, and any subsequent results simulated will appear under that first config, even as the last appears to be selected. This should be a relatively straightforward frontend fix with just updating the correct shading for the selected config after loading the model.

Issue #269, opened 2021-02-01
Naming functionality for configurations and results


esterzhao:
* Allow users to add names to the configurations and results

Issue #265, opened 2021-01-27
Streamline naming convention to map HTML elements to analysisMap


leshghi:
Currently naming convention for HTML elements to connect them to analysisMap configs and results is a bit chaotic. Need a more code-reading friendly/efficient way to connect the two, or a set convention. Part of the reason this is choppy and messy right now is due to the future implementation of renaming configs and results, so if you make any changes to set a new convention for naming/mapping please keep that in mind as well.

PR #283, opened 2021-02-17
Naming feature

leshghi:
This branch primarily adds a naming feature for the configurations, and updates our html naming conventions and event listeners to primarily tie directly to the html elements rather than through id strings. This should close Issue #269 and #265. 

I know Issue #269 originally included updating result names as well, and this is absolutely still possible, but would involve some space and complexity tradeoffs. Basically it would change the structure of either the AnalysisResults object itself, or the analysisResults param in our AnalysisConfig object, so I think we should have a bigger discussion about the use cases for naming results, and whether we still want to implement that feature. 

Incidentally, this branch should also close Issue #282, and #262.

Because this branch involved so many changes and kind of overhauled our html structure, I am sure there are a LOT of bugs. I've done my best searching for and fixing as many as I can, but just wanted to say in advance that this is Super far from a polished (more like jagged and dull) pull request. But I've hit a wall with finding bugs and so I'm hoping for as many eyes as possible to help me find more so that we can merge this in ASAP because it touches almost all of our model-management code I'm pretty sure and will cause a super fun amount of merge conflicts. Okay thats all thank you in advance!

### DIFF
--- a/leaf-ui/js/displayAnalysis.js
+++ b/leaf-ui/js/displayAnalysis.js
@@ -267,7 +267,7 @@ function loadAnalysis(){
         // Add the config to the sidebar
         addAnalysisConfig();
         // Add the results (if any) to the sidebar
-        updateResults();
+        loadResults();
         console.log(analysisRequest);
     }
     // TODO: figure out how to set it to the element of the map that will populate on top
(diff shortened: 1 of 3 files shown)


krhablutzel:
This is a lot of changes! Thank you for making all these adjustments! I'll start a bug hunt. Looking forward to the massive merge conflict :)

ibrodycalixto:
2 bugs that I found so far:
- The box does not go back to a non-input box after clicking away
- Names are not saved (when you save and load named configs, default naming used)
I'll keep looking!

leshghi:
> 2 bugs that I found so far:
> 
> * The box does not go back to a non-input box after clicking away
> * Names are not saved (when you save and load named configs, default naming used)
>   I'll keep looking!

Ah yes you have to press the enter key to save the names! Which is probably why they are not saving. Do you think clicking away to save is more intuitive? I can def try to make that update either here or in another issue/bug fix.

ibrodycalixto:
> > 2 bugs that I found so far:
> > 
> > * The box does not go back to a non-input box after clicking away
> > * Names are not saved (when you save and load named configs, default naming used)
> >   I'll keep looking!
> 
> Ah yes you have to press the enter key to save the names! Which is probably why they are not saving. Do you think clicking away to save is more intuitive? I can def try to make that update either here or in another update.

Lol I was just about to comment to say that I read the code and understand now. Clicking away is my first instinct, but we can ask in lab!

ibrodycalixto:
> > > 2 bugs that I found so far:
> > > 
> > > * The box does not go back to a non-input box after clicking away
> > > * Names are not saved (when you save and load named configs, default naming used)
> > >   I'll keep looking!
> > 
> > 
> > Ah yes you have to press the enter key to save the names! Which is probably why they are not saving. Do you think clicking away to save is more intuitive? I can def try to make that update either here or in another update.
> 
> Lol I was just about to comment to say that I read the code and understand now. Clicking away is my first instinct, but we can ask in lab!

This is also why names didn't save...my bad teehee. But I did find a real bug!

- When they save, the named configs are loaded last, unnamed first. So loading named configs puts them in the wrong order

leaf-ui/js/displayAnalysis.js
```diff
@@ -305,24 +303,24 @@ function addNewAnalysisConfig(){
...
-        + '</div><div class = "dropdown-container" id="'+label+'"></div>'
-        + '</div>'
-      );
+
+    // Add config to config container
+    $("#configurations").append(getConfigHtml(config.id));
+
+    // Get main config element and add event listeners for all interactive components
+    mainElement = document.getElementById(config.id);
+    mainElement.querySelector('.log-elements').addEventListener('dblclick', function(){rename(this /** Config element */)});
```
krhablutzel: What’s the origin of the class name `log-elements` for the buttons with config names? I'm not sure it's clear these are the config buttons based on the naming, but I might be missing something.

leshghi:
Updated so that clicking outside input box also saves name!

krhablutzel (approved):
My branch is merged, so this one should be set for now! (Apologies for the merge conflicts)

leaf-ui/js/displayAnalysis.js
leshghi: Updated to better name!

PR #291, opened 2021-03-08
Removes all history log code 

mvarnum:


leaf-ui/index.html
```diff
@@ -159,7 +166,12 @@ <h3 style="text-align:left; color:#1E85F7; margin-bottom:5px;">Relative Intentio
         
 
         <div id="stencil" class="left-panel"><label>Stencil</label></div>
-        <div id="history" class="left-panel"><label>History Log</label></div>
+        <!-- <div id="analysis-sidebar" class="left-panel"><h3 style="text-align:left; color:#181b1fe3; margin-bottom:5px; margin-left: 10px;">Analysis
```
krhablutzel: I think the analysis sidebar needs to be uncommented now

leaf-ui/js/object/InputAnalysis.js
```diff
@@ -12,7 +12,7 @@ class InputAnalysis {
 		this.initialAssignedEpoch = savedAnalysisData.finalAssignedEpoch ? savedAnalysisData.finalAssignedEpoch : ['0'];
 		this.initialValueTimePoints = savedAnalysisData.finalValueTimePoints ? savedAnalysisData.finalValueTimePoints:
 			['0'];
-		this.elementList = getElementsForAnalysis();
+		//this.elementList = getElementsForAnalysis();
```
krhablutzel: Is the InputAnalysis class even used anymore? I know it mirrors an InputAnalysis class in the backend, but I couldn't find it being used anywhere in the frontend. Wondering if we even need to have it

leaf-ui/js/object/InputAnalysis.js
```diff
@@ -73,24 +73,24 @@ function getElementList() {
...
-		var index = parseInt(time) + 1;
-		historyObject.allHistory[historyIndex].analysis.elements.forEach(function(currentElement) {
-				var element = {};
-				element.id = currentElement.id;
-				element.status = currentElement.status.slice(0, index);
-				elementList.push(element);
-			}
-		);
-	}
+// function getElementsForAnalysis() {
```
krhablutzel: Not sure whether we need to leave this... probably related to whether we still need the InputAnalysis class?

krhablutzel (changes requested):
Thanks for making all these changes! I think the analysis sidebar needs to be uncommented so it works again, but otherwise looks great. A lot of changes have been made to model-management since this branch was created, so be prepared for possible merge conflicts. Not quite approving this yet since I'd love to test that the analysis sidebar still works, but overall looks very good!

krhablutzel (approved):
This looks great Megan! Thanks for making all these changes; you can probably merge now!

esterzhao (approved):
Looks good!

PR #298, opened 2021-03-15
Fixed bug that fails to properly update initial satisfaction values when returning to modeling mode

### LINKED ISSUE #281: Satisfaction values on intentions don't clear when switching to modeling view
This appears to happen both on model-management and develop, so more than a model-management issue, but might as well address it now.
After running an analysis, the satisfaction values don't disappear from intentions when we return to modeling mode.
Screen Shot 2021-02-15 at 16 24 19

mvarnum:


mvarnum:
Addressed bug #281

krhablutzel (commented):
This looks great to me. Thanks for making these changes!

krhablutzel (approved):
Looks great, you have my approval.

PR #302, opened 2021-03-16
Analysis object refactor

YanningTan:
In backEndComm.js, the displayAnalysis will take analysisResult rather than raw results. The functionality of objectAnalysis has been moved to class displayAnalysis().  functions of objectAnalysis nextStates and initFromMain has been deleted since they were not called. An instance variable this.type has been deleted.


leaf-ui/js/backendComm.js
```diff
@@ -175,7 +175,8 @@ function getFileResults(isGetNextSteps){
                     analysisRequest.previousAnalysis = analysisResult;
                     console.log("previousAnalysis");
 					console.log(analysisRequest.previousAnalysis);
-					displayAnalysis(results);
+                    //display analysis result rather than raw results
```
esterzhao: I appreciate the comment! For posterity, can we duplicate this comment on line 103 where `displayAnalysis()` is also called, and maybe instead say `// pass displayAnalysis an AnalysisResult object`?

leaf-ui/js/displayAnalysis.js
```diff
@@ -19,7 +19,8 @@ var currAnalysisConfig;
 function displayAnalysis(analysisResults){
 
     // Change the format of the analysis result from the back end
-    var currentAnalysis = new analysisObject.initFromBackEnd(analysisResults);
+    var currentAnalysis = new AnalysisResult();
```
esterzhao: My hope is that since in backendComm.js we are passing in the `AnalysisResult` object (defined in Class.js) that we should not be creating a new `AnalysisResult` obj here. Ideally, we could just say `var currentAnalysis = analysisResults` and then call the `initFromBackEnd` function something more accurate that describes what the variables you are creating there will be used by.

leaf-ui/js/object/Class.js
```diff
@@ -343,7 +343,36 @@ class AnalysisResult {
         this.colorVis; //color visualization for analysis mode  
         this.selectedTimePoint; //find where slider is initialized and set timepoint in here. Also place it in update function
         this.maxTimePoint;
+
+
+        this.elements;
```
esterzhao: Is there any particular reason that we should have two variables that hold the elements, `elementList` and `elements`? I also think `numOfElements` is redundant when we can simply do `currAnalysis.elements.size()` or something like that. It just adds another variables that needs to be updated when elements are added, which I think is overly complicated.

leaf-ui/js/object/Class.js
```diff
@@ -343,7 +343,36 @@ class AnalysisResult {
...
+        this.numOfElements = Number(analysisResults.elementList.length);
+        this.timeScale = Number(analysisResults.timePointPath.length) - 1;
+        this.relativeTime = [];
+
+        for (var i = 0; i < analysisResults.timePointPath.length; i++) {
+            var aux = analysisResults.timePointPath[i];
+            this.relativeTime.push(aux);
+        }
+
+        for (var i = 0; i < this.numOfElements; i++) {
```
esterzhao: As stated above, this seems unnecessary - you're basically just copying the full element list that exists already onto ANOTHER class variable. Instead, try to go through the code in displayAnalysis and rework it to use the `elementList` variable instead. Unless theres a specific reason to run this for-loop!

leaf-ui/js/object/Class.js
```diff
@@ -343,7 +343,36 @@ class AnalysisResult {
...
+        this.relativeTime;
     }
+    // Parse results from backend
+    initFromBackEnd = function (analysisResults) {
+        this.elements = [];
+        this.numOfElements = Number(analysisResults.elementList.length);
+        this.timeScale = Number(analysisResults.timePointPath.length) - 1;
+        this.relativeTime = [];
+
+        for (var i = 0; i < analysisResults.timePointPath.length; i++) {
```
esterzhao: it doesn't look like you are doing anything significant to this variable either - would it just be a copy of the already existing `timePointPath` variable? Because in that case, just rework the other code to accept the `timePointPath` variable instead of `relativeTime`

leaf-ui/js/object/Class.js
```diff
@@ -343,7 +343,36 @@ class AnalysisResult {
...
         this.maxTimePoint;
+
+
+        this.elements;
+        this.numOfElements;
+        this.timeScale;
+        this.relativeTime;
     }
+    // Parse results from backend
+    initFromBackEnd = function (analysisResults) {
```
esterzhao: A couple things here - first up, can you rework this so that it's written in the same format as the constructor? The purpose of this rework is to move away from older JS class structures that don't look as nice! 
ex:
`initFromBackEnd() {  code here }`

I see that you call this method from displayAnalysis by passing the AnalysisResult object from the backend in. As I stated above, I'd prefer to just use the same object instead of copying everything over onto another one! Seems like a bit of a waste of time. So since this is a class method, we would not be accepting the `AnalysisResult` as an input -- Instead, if you need to reference the `AnalysisResult` object, do so using `this`.

esterzhao (changes requested):
Ok so a few things to fix here, which are partly my fault because I wasn't clear enough on what I wanted from you! Let me know if you want to meet to discuss what these mean.

leaf-ui/js/object/Class.js
```diff
@@ -338,12 +346,26 @@ class AnalysisResult {
         this.timePointPathSize;
         this.elementList; 
         this.allSolution;
-        this.elementListPercentEvals;
         this.isPathSim = false; //used for slider visualization
         this.colorVis; //color visualization for analysis mode  
         this.selectedTimePoint; //find where slider is initialized and set timepoint in here. Also place it in update function
-        this.maxTimePoint;
+        this.timeScale;
+
```
esterzhao: Could you delete some of this white space? Just for prettiness

leaf-ui/js/object/Class.js
```diff
@@ -319,7 +319,10 @@ Actor.numOfCreatedInstances = 0;
 
 class AnalysisResult {
     /**
+     *
      * @param {Array.<String>} assignedEpoch
+     *   Each element represents an epoch with its assigned value
```
esterzhao, line 323: Thanks so much for adding these params and explanations 🥰

leaf-ui/js/object/Class.js
```diff
@@ -338,12 +346,26 @@ class AnalysisResult {
         this.timePointPathSize;
         this.elementList; 
         this.allSolution;
```
esterzhao: I'm guessing this is the allSolution param you mentioned in your message -- I'll do some testing but for now we can just leave it there, could you maybe add a comment on it that just says like `// potentially deprecated` or something like that?

leaf-ui/js/object/Class.js
```diff
@@ -338,12 +346,26 @@ class AnalysisResult {
...
-        this.elementListPercentEvals;
         this.isPathSim = false; //used for slider visualization
```
esterzhao: Could you add @params for these variables too?

leaf-ui/js/object/Class.js
```diff
@@ -338,12 +346,26 @@ class AnalysisResult {
...
+        
+
+
+
+
+
     }
+    getTimeScale() {
+        this.timeScale = Number(this.timePointPath.length) - 1;
+        return this;
```
esterzhao: You shouldn't have to `return this` -- that would return the AnalysisResult object you are already working on, which won't be necessary

leaf-ui/rappid-extensions/AnalysisInspector.js
```diff
@@ -108,7 +108,7 @@ var AnalysisInspector = Backbone.View.extend({
 	 */
 	getAllNextStates: function () {
 		if(analysisRequest.action != null) { //path has been simulated
-			if(analysisResult.selectedTimePoint != analysisResult.maxTimePoint) { //last timepoint is not selected
+			if (analysisResult.selectedTimePoint != analysisResult.timeScale) { //last timepoint is not selected
```
esterzhao, line 111: good catch for this replacement!

leaf-ui/js/displayAnalysis.js
```diff
@@ -19,7 +19,8 @@ var currAnalysisConfig;
 function displayAnalysis(analysisResults){
```
esterzhao: Could you update the docstring of this to specify `@param {AnalysisResult} analysisResult`? That way it makes it clear that this function is being passed an AnalysisResult object

esterzhao: same should be true for any other method that is taking in the currentAnalysis object as a param - maybe walk through displayAnalysis.js and update all docstrings

esterzhao (changes requested):
Pull changes from the model-management branch, make a few changes, and then you will be good to go! Great job @YanningTan :)

PR #306, opened 2021-03-19
Clear results functionality

### LINKED ISSUE #305: Clear results functionality
Allow the user to clear results, or results + configs, as two separate clear options

krhablutzel:
Added a clear option which removes just results, not configurations, in addition to the clear option which removes both results and configurations. Adapted the clear dropdown so that these two analysis clearing options appear in analysis mode, and other clearing options appear in modeling mode. Resolves #305

### DIFF
--- a/leaf-ui/index.html
+++ b/leaf-ui/index.html
@@ -42,10 +42,11 @@
                 <!-- <div class="btn">Font Size</div> -->
                 <div class="dropdown-toolbar">
                   <a id="btn-clear-all" class='model-clears'>Full Model and Analysis</a>
-                  <a id="btn-clear-analysis">Analysis</a>
                   <a id="btn-clear-elabel" class='model-clears'>Evaluation Labels</a>
                   <a id="btn-clear-flabel" class='model-clears'>Dynamic Function Labels</a>
                   <a id="btn-clear-cycle" class='model-clears'>Cycle Highlighting</a>
+                  <a id="btn-clear-analysis" class='analysis-clears' style="display: none;">Configurations and Results</a>
+                  <a id="btn-clear-results" class='analysis-clears' style="display: none;">Results</a>
                 </div>
             </div>
             <button id="btn-zoom-in" class="btn" onclick = "zoomIn(paperScroller)">Zoom In</button>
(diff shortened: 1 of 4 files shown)


leaf-ui/js/displayAnalysis.js
```diff
@@ -330,6 +330,19 @@ function updateResults(){
...
 
+/**
+ * Clear all results from all configs
+ */
+ function clearResults(){
+    for(let config of analysisMap.values()) {
+        // remove analysis results from each config
+        config.deleteResults();
+        // remove all results from config div
+        var dropdownElement = document.getElementById(config.id).querySelector('.dropdown-container');
```
leshghi: 100% optional: You could use "$('.result-elements').remove();" here outside of the for-loop if you wanted to be (very slightly) more concise, but if you'd rather not use JQuery here understood

leaf-ui/index.html
```diff
@@ -42,10 +42,11 @@
                 <!-- <div class="btn">Font Size</div> -->
                 <div class="dropdown-toolbar">
                   <a id="btn-clear-all" class='model-clears'>Full Model and Analysis</a>
-                  <a id="btn-clear-analysis">Analysis</a>
                   <a id="btn-clear-elabel" class='model-clears'>Evaluation Labels</a>
                   <a id="btn-clear-flabel" class='model-clears'>Dynamic Function Labels</a>
                   <a id="btn-clear-cycle" class='model-clears'>Cycle Highlighting</a>
+                  <a id="btn-clear-analysis" class='analysis-clears' style="display: none;">Configurations and Results</a>
+                  <a id="btn-clear-results" class='analysis-clears' style="display: none;">Results</a>
```
leshghi: Super minor, but in keeping with the style of the save dropdown would we want to put results before configurations and results? Or maybe just say "Results only"? Idk I may just be weird here but something abt how this reads compared to our other dropdowns just feels slightly off to me. I don't really know a better way though so maybe ignore this and if other people end up feeling similarly or a new phrasing/order comes to us we can potentially make a change then.

leshghi (approved):
Amazing job 10/10 I left one super tiny optional suggestion and one musing comment!

leaf-ui/js/displayAnalysis.js
krhablutzel: Love this idea, I'll make that change!

leaf-ui/index.html
krhablutzel: This is a great point. I've changed it to "Results Only" on top and "Configurations and Results" on the bottom, but I think we eventually need a broader conversation about the style of our dropdowns. In save, each option is "Save ____" while Clear doesn't preface everything with "Clear," and there's also the question of calling them "Analysis Configurations" versus just "Configurations"

PR #308, opened 2021-03-19
Slider starts at 0 when switching results

### LINKED ISSUE #307: Slider should start at 0 when switching to existing results
In the history log, the slider started at 0 when switching to old results. The slider only started at the max time when results were first created. With the new model management features, the slider should start at 0 whenever the user switches to an already existing result.

krhablutzel:
Closes #307 such that the slider starts at 0 when switching to existing results and starts at max time when displaying new results from the backend.


esterzhao (approved):
Great fix Kathleen! Super clean.

PR #309, opened 2021-03-19
Add alert during rename when name is taken

leshghi:
Adds popup when name is already taken instead of just console logging so that the user knows why the name is not updating.


krhablutzel (approved):
10/10 looks great!

PR #310, opened 2021-03-22
Iss294 Model Editing Features

### LINKED ISSUE #294: Handle model editing after creation of analysis configurations
* Decide on implementation - warning? update configs?
* If one node is deleted, do all configs disappear?

krhablutzel:
Clears results when the model is changed, and warns users that changing the model will clear results. Updates configurations with the changed model. Detecting when the model is changed is still imperfect due to duplication of nodeIDs in the Actor's intentionID list, which is an outstanding bug.

Closes #294

### DIFF
--- a/leaf-ui/js/loadSaveFunctions.js
+++ b/leaf-ui/js/loadSaveFunctions.js
@@ -264,6 +264,7 @@ function loadFromObject(obj) {
 	model.maxAbsTime = obj.model.maxAbsTime;
 
 	// store deep copy of model for detecting model changes
+	// copy is NOT of type Model
     previousModel = JSON.parse(JSON.stringify(model));
 
 	// Clear any previous analysis data 
--- a/leaf-ui/js/loadSaveFunctions.js
+++ b/leaf-ui/js/loadSaveFunctions.js
@@ -263,6 +263,9 @@ function loadFromObject(obj) {
 	model.constraints = getConstArr(obj.model.constraints);
 	model.maxAbsTime = obj.model.maxAbsTime;
 
+	// store deep copy of model for detecting model changes
+    previousModel = JSON.parse(JSON.stringify(model));
+
 	// Clear any previous analysis data 
 	if (analysisMap.size != 0) {
 		// Clear the analysisMap to remove any previous analysis
--- a/leaf-ui/js/objects.js
+++ b/leaf-ui/js/objects.js
@@ -176,6 +176,7 @@ class ChartObj {
 	}
 }
 
+// TODO: Clean up this config. We potentially only need the id, analysisRequest, and analysisResults params
 class AnalysisConfiguration {
 	/**
 	 * This class is used to hold analysis configuration specifications and results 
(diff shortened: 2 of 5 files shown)


leaf-ui/js/objects.js
```diff
@@ -176,6 +176,7 @@ class ChartObj {
 	}
 }
 
+// TODO: Clean up this config. We potentially only need the id, analysisRequest, and analysisResults params
```
esterzhao, line 179: You're probably right! All the stuff is duplicated from the analysisRequest, only thing we'd have to account for is the slight confusing involving analysisRequests and loading in files, etc

esterzhao:
I'm currently doing a review, but I wanted to leave a comment on this -- I've found that I can actually move goals and tasks around while in Analysis mode, which can affect the model when you move them outside of actors. I'm not sure how the analysis is registering these changes, but if you move them out of actors I would classify that as a significant change, and it might be better if we restricted any change to the model in Analysis mode. maybe not for this branch, but to keep in mind and related

leaf-ui/js/loadSaveFunctions.js
```diff
@@ -263,6 +263,10 @@ function loadFromObject(obj) {
 	model.constraints = getConstArr(obj.model.constraints);
 	model.maxAbsTime = obj.model.maxAbsTime;
 
+	// store deep copy of model for detecting model changes
```
esterzhao, line 266: Not sure where to put this comment, but I found a bug, I'm not sure what is causing it. If I use the bikebefore.json test file from our repos this functionality doesn't work. I'm not sure if thats because there are no actors or something strange like that, but worth taking a look into

leaf-ui/js/objects.js
```diff
@@ -255,6 +256,14 @@ class AnalysisConfiguration {
 		this.id = id;
 	}
 
+	/**
+	 * Updates user assignments list param for config and for config's analysisRequest
+	 */
+	updateUal(userAssignmentsList){
```
esterzhao: nit, can we rename to `updateUAL`? Just bc its an acronym

leaf-ui/js/onFunctions.js
```diff
@@ -522,6 +542,26 @@ function switchToModellingMode() {
 	graph.allElements = null;
     mode = "Modelling";
     EVO.switchToModelingMode();
+
+    // Popup to warn user that changing model will clear results
```
esterzhao, line 546: This popup is great work! Thanks guys :)

esterzhao (approved):
I left a few comments with nitpicks as well as a call out of a potential bug, but ultimately this is an approve! I did some very annoying testing as a really dumb user, and everything other than that one bug seems to be working. See my comment about making changes from Analysis mode, and maybe we can discuss in lab on Monday

leaf-ui/js/loadSaveFunctions.js
krhablutzel: Would love to chat more about this bug, because I haven't been able to recreate it!

PR #322, opened 2021-03-30
Scroll bar

### LINKED ISSUE #315: Configs are being covered by the watermark and cannot be accessed
Screen Shot 2021-03-26 at 11 30 58 AM
* Need to either add a scroll bar or make the watermark its own div that cannot overlay the config sidebar div

leshghi:
Here I reduced the size of the configuration div and added a scroll bar to make it easier to access configs that were previously covered by the watermark. I also moved the actor element up in modelling view to minimize the risk of the watermark overlapping as well. However, these are both temporary fixes. A larger CSS fix needs to be done to separate the watermark from the left panel div and figure out how to best do that. This may be slightly more difficult in modelling mode, as well just in terms of consistency between two modes and the way the joint.js theme and css works for the node element panel. Pushing that off to our aesthetic UI/CSS overhaul with @krhablutzel (thanks Kathleen!), but this patch fix should at least lower the chances of inaccessible configs in the meantime. This was a super fast and dirty fix so any improvements/suggestions/other paths for this are more than welcome!

### DIFF
--- a/leaf-ui/index.html
+++ b/leaf-ui/index.html
@@ -172,7 +172,7 @@
             <div class="addConfig" style="display:inline">
                 <i class="fa fa-plus" id="addIntent" style="font-size:30px; float:right;  margin-bottom:5px; margin-right:20px;"></i>
             </div></h3>
-            <div id="configurations" class="left-panel" style="margin-top:20px; overflow-y:auto; height:69%"></div>
+            <div id="configurations" class="left-panel" style="margin-top:20px; overflow-y:auto; height:69%; box-shadow: none;"></div>
         </div>
 
         <div id="paper">
--- a/leaf-ui/index.html
+++ b/leaf-ui/index.html
@@ -166,14 +166,13 @@
 		      <button id="symbolic-btn" class="blue-btn btn">Model Constraints</button>-->
             <button id="analysis-btn" class="green-btn btn" style="border-radius:none;">Analysis</button>
         </div>
-        
-
+         
         <div id="stencil" class="left-panel"><label>Stencil</label></div>
         <div id="analysis-sidebar" class="left-panel"><h3 style="text-align:left; color:#181b1fe3; margin-bottom:5px; margin-left: 10px;">Analysis
             <div class="addConfig" style="display:inline">
                 <i class="fa fa-plus" id="addIntent" style="font-size:30px; float:right;  margin-bottom:5px; margin-right:20px;"></i>
             </div></h3>
-            <div id="configurations" class="left-panel" style="margin-top:20px"></div>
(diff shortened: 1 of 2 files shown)


leaf-ui/index.html
```diff
@@ -166,14 +166,13 @@ <h3 style="text-align:left; color:#1E85F7; margin-bottom:5px;">Relative Intentio
...
-        
-
+         
         <div id="stencil" class="left-panel"><label>Stencil</label></div>
         <div id="analysis-sidebar" class="left-panel"><h3 style="text-align:left; color:#181b1fe3; margin-bottom:5px; margin-left: 10px;">Analysis
             <div class="addConfig" style="display:inline">
                 <i class="fa fa-plus" id="addIntent" style="font-size:30px; float:right;  margin-bottom:5px; margin-right:20px;"></i>
             </div></h3>
-            <div id="configurations" class="left-panel" style="margin-top:20px"></div>
+            <div id="configurations" class="left-panel" style="margin-top:20px; overflow-y:auto; height:69%"></div>
```
krhablutzel: Can we add `box-shadow: none;` to the style? The shadow is a bit odd in the middle of the left panel.

krhablutzel: I do see how it could be useful to have some line to indicate the bottom of the scroll box when we add enough results for the results to go off the bottom of the scroll bar, but maybe there's a way to auto-scroll to where the latest result was added? Outside the scope of this pull request, of course.

krhablutzel (approved):
This is a great temporary fix! I left one suggestion for removing the box shadow in the middle of the left sidebar, and some musings on future CSS fixes, but otherwise this looks like a great patch for now.

Notes for the future, that just happened to notice but don't need to be fixed now:
- I can still scroll down because the analysis inspector panel is too long.
- The watermark will still overlap the scroll box/the actors when the webpage gets shorter.
- We need to update the watermark with Smith info
- Could we auto-scroll to where new results are added? Or else add a less shadow-y line at the bottom of the scroll box so the user can see it's scrolling.

leaf-ui/index.html
leshghi: Yes I can absolutely do that! I personally like the shadow as a separation of divs, but it definitely looks a bit wonky especially when the modelling mode left side panel doesn't have the same thing, so I'll change it for this request! I'm just waiting for Esters comments to make final changes and push a commit. 

I think in a future world we have to figure out how to handle the watermark consistently across modes - you hit on a bunch of much larger issues that I also noticed but couldn't quite figure out how to fix at least without some bigger changes. This is where you come in Kathleen! Tbh if you as a css whiz have anything you want to add to this pull request you can def do it directly on the branch, but obviously no pressure at all!

PR #323, opened 2021-03-30
Reset graph UI when clearing results/configs

### LINKED ISSUE #321: Clearing results/configs does not clear currently selected result in UI/on graph

leshghi:
Fixes #321. I did this by moving the removeSlider() function into the revertNodeValuesToInitial() function, and calling it inside the functiosn triggered by clear results and/or configs. I looked at use cases for revertNodeValuesToInitial() and noticed that it was always called in tandem with removeSlider() so thought it would be more efficient to move it directly inside that function.


krhablutzel (approved):
10/10 this all looks fantastic! Thanks for cleaning this up!

PR #324, opened 2021-03-30
Correctly update mode tag in modeling view

### LINKED ISSUE #320: When switching back to Modeling View, "Analysis" tag remains unchanged
This error exists on develop branch as well, so a fix could be a branch off of develop. 

Should change back to "Model View" text when switching back

Screen Shot 2021-03-30 at 2 02 38 PM

leshghi:
Also updated the analysis view tag to reflect the original styling of the model view tag -> aka "Analysis View" to match "Modeling View" instead of just "Analysis." Fixes #320.

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -410,7 +410,7 @@ function switchToAnalysisMode() {
 
     $('#model-toolbar').css("display", "none");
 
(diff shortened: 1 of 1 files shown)


esterzhao (approved):
Thanks @leshghi ! What a simple fix, can't believe its just happening now haha

PR #325, opened 2021-04-02
Fixes 3 outstanding issues from develop branch - #255, #304, #314

### LINKED ISSUE #255: Taskbar formatting fix - EVO
In Analysis view, once Simulate Single Path is run, the EVO label overlays View Assignments List button and makes it difficult to fix. Should just be a formatting fix for the frontend. 
evo_glitch

### LINKED ISSUE #304: Right side modeling panel throws errors when switching initial satisfaction value
When the initial satisfaction value is not "no value" and you switch from modeling to analysis more and then back, if you try to switch it to no value again, it throws an error. In that case "no value" should either not be a selection on the menu, or should have a way to handle that error.

### LINKED ISSUE #314: Analysis view allows user to edit configuration without deleting existing results
* needs to include a similar popup to model editing, or auto delete any results
* Also documented in #233

leshghi:
Fixes #255, fixes #304, fixes #314 - all relatively quick few line fixes.

Note that there is an outstanding model loading issue in development but not occurring/presumably fixed in the model-management branch which means the initial sat values do not work properly if the model is loaded. #304 does not fix this issue as it should be corrected when model-management is merged.

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -1033,7 +1033,7 @@ paper.on('blank:pointerdown', function(evt, x, y) {
 });
 
 /**
- * TODO: Unclear if this is actually doing anything - look into possible deletion
+ *
  */
 paper.on('cell:pointerdown', function(cellView, evt, x, y) {
 	
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -389,6 +389,7 @@ function reassignIntentionIDs() {
  */
 var inAnalysis = false;
 function switchToAnalysisMode() {
+    setInteraction(false);
     inAnalysis = true;
 	reassignIntentionIDs();
 	
@@ -475,6 +476,7 @@ function revertNodeValuesToInitial() {
  * Display the modeling mode page
  */
 function switchToModellingMode() {
+    setInteraction(true);
     analysisResult.isPathSim = false; //reset isPathSim for color visualization slider
 	analysisRequest.previousAnalysis = null;
     clearInspector();
@@ -1031,7 +1033,7 @@ paper.on('blank:pointerdown', function(evt, x, y) {
 });
 
 /**
- * 
+ * TODO: Unclear if this is actually doing anything - look into possible deletion
  */
 paper.on('cell:pointerdown', function(cellView, evt, x, y) {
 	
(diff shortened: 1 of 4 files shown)


esterzhao:
Notes for me before I do my full review:
#255 fixed, is there a possibility for interference with @krhablutzel 's fix for the order of the things in the top bar?
#304 I don't super understand so someone else might need to test it to make sure, but looks like it was fixed!
#314 really important fix, thanks!

leaf-ui/js/onFunctions.js
```diff
@@ -1031,7 +1033,7 @@ paper.on('blank:pointerdown', function(evt, x, y) {
 });
 
 /**
- * 
+ * TODO: Unclear if this is actually doing anything - look into possible deletion
```
esterzhao: I'd rather not leave TODOs in develop if we can avoid it. Maybe @amgrubb can chime in?

leaf-ui/js/onFunctions.js
```diff
@@ -1319,3 +1321,14 @@ function checkForMultipleNB(node) {
 	return num >= 1;
 }
 
+/**
+ * Sets interaction option on all elements in graph
+ * 
+ * @param {boolean} interactionValue 
+ */
+ function setInteraction(interactionValue){
```
esterzhao, line 1329: so simple .... so clean ... thanks bud

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -74,7 +74,7 @@ var ElementInspector = Backbone.View.extend({
             '<textarea class="cell-attrs-text"></textarea>',
             '<label>Initial Satisfaction Value:</label>',
             '<select id="init-sat-value">',
-                '<option value="(no value)"> (no value) </option>',
+                '<option value="(no value)" disabled> (no value) </option>',
```
esterzhao, line 77: again, so simple. great fix!

esterzhao (approved):
Want a few other people to take a speedy quick look, so I've tagged them and assigned above! See my comment for more info. Otherwise, looks really good, I tested to the best of my ability and everything looks great. Thanks Ali!

krhablutzel (approved):
These all look great to me. Re: Ester's concern, I don't see any potential interference with the order changes I made in the top bar in `model-management`, so not expecting any merge conflicts when we eventually merge `model-management` into `develop`. But thanks for checking, Ester!

These are all very elegant solutions, Ali—great work!

PR #330, opened 2021-04-12
added spinner under cursor for explore next states

### LINKED ISSUE #236: Add Spinner for Explore Next States feature
Since Explore Next States feature takes >5 seconds to load on many computers - add pop-up or spinner so that user knows it is in process and the page has not just frozen/broken

YesugenB:
Fixes #236 
Added spinner under cursor for when loading Explore Possible Next States and Explore Next States pop-up window.

### DIFF
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -17,8 +17,6 @@ body {
   font-family: Arial;
 }
 
-body.waiting,
-body.waiting a:hover,
 body.waiting *
 {
     cursor: wait !important;
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -17,6 +17,13 @@ body {
   font-family: Arial;
 }
 
+body.waiting,
+body.waiting a:hover,
+body.waiting *
+{
+    cursor: wait !important;
+}
+
 .header {
    padding: 0px;
 }
--- a/leaf-ui/css/style.css
+++ b/leaf-ui/css/style.css
@@ -17,8 +17,6 @@ body {
   font-family: Arial;
 }
 
-body.waiting,
-body.waiting a:hover,
 body.waiting *
 {
     cursor: wait !important;
--- a/leaf-ui/css/style.css
+++ b/leaf-ui/css/style.css
@@ -17,6 +17,13 @@ body {
   font-family: Arial;
 }
 
+body.waiting,
+body.waiting a:hover,
+body.waiting *
+{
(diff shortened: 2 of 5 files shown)


leaf-ui/css/analysis_html.css
```diff
@@ -17,6 +17,13 @@ body {
   font-family: Arial;
 }
 
+body.waiting,
```
leshghi: I'm not sure if you need the body.waiting, body.waiting a:hover since you have body.waiting * but I am not a css whiz and could be totally off base so feel free to correct me!

leaf-ui/js/analysis.js
```diff
@@ -1404,7 +1404,8 @@ function save_current_state(){
 
 //This function should get the current state and generate a new window with the next possible states
 function generate_next_states(){
-
+
```
leshghi, line 1407: Would be great if you could add a comment to this line

leaf-ui/rappid-extensions/AnalysisInspector.js
```diff
@@ -150,7 +151,7 @@ var AnalysisInspector = Backbone.View.extend({
 			swal("Error: Cannot explore next states before simulating a single path.", "", "error");
 		}
 
-
+
```
leshghi: nit: if you could delete this extra line that would be amazing

leaf-ui/css/style.css
```diff
@@ -17,6 +17,13 @@ body {
   font-family: Arial;
 }
 
+body.waiting,
```
leshghi: same q here around body.waiting & body.waiting a:hover

leshghi (approved):
Yesugen this looks amazing and the user experience is SO improved - thank you for single-handedly taking on the next states window/feature. I left a few super minor comments which are all just stylistic. 

I also wanted to note 2 potential "new issues" that came up while I was testing which may or may not be fixable but we could probably document for the future. The first is that the user can technically still click buttons while the next states window is loading - so maybe eventually removing that ability would be good if possible. The other is that there is often a large delay between selecting a filter and the checkbox actually appearing as clicked - I did some initial research and this lag is because we call a logic/computation heavy function onclick(), so it performs that before rendering the click in the UI, but I think there should either be a spinner while it computes as well or it should update ui and then perform the computations.

Anyway this pull request has my enthusiastic approval - thank you again!

leaf-ui/css/analysis_html.css
YesugenB: I think you're right!! just changed it

PR #343, opened 2021-04-25
Resolves bugs with interaction of actors and intentions

### LINKED ISSUE #312: Actors store duplicate nodeIDs
When BloomingLeaf thinks we're dragging a node (which seems to be every time the node is clicked), the node is removed from the Actor and then re-added when we stop clicking. However, the nodeID is never removed from the Actor's list of intentionsIDs, resulting in duplicate nodeIDs getting added to the Actor's intentionsIDs list every time the intention is clicked.

krhablutzel:
This branch resolves multiple bugs and cleans up code related to the interaction of actors and intentions:

- actors stored duplicate nodeIDs from every time an intention was re-added to the actor — now such duplication is prevented
- actors did not remove nodeIDs from their intentionID lists when intentions were dragged out of the actor — fixed this removal process
- intentions could not reparent to no actor (`'-'`) when dragged out of all actors — added this functionality
- intentions were removed and re-added to the actor on every click — now they are only removed and re-added when dragged; same with reparenting links
- removing intentions from actors also removed the intentions from the userAssignmentsList (UAL), even if the intention wasn't removed from the model — this duplicate and sometimes incorrect functionality was deleted because other existing code already removes intentions from the UAL in all proper places
- `cell.on('remove'...)` behavior did not run for intentions, links, and actors loaded from a saved model — these functions were deleted because the exact same behavior exists in `paper.on('remove'...)` and works for both loaded and non-loaded models

In addition, these changes restructure `'cell:pointerdown'`,  `'cell:pointermove'`, and `'cell:pointerup'`events to distinguish between clicking and dragging an intention/link, such that intentions/links only get re-embedded/reparented when moved rather than every time the user clicks them. There are also small changes to simplify the logic flow in these events, such as not checking whether a cell is a Link inside an `else` statement that only runs when the cell is not a Link.

All these changes are intended to address the bug in `model-management` where clicking on an intention changes the model, since every time the intention was clicked, it would be removed from the actor below (without actually being removed, due to a bug) and then re-embedded in the actor (with duplication allowed in the actor's intentionsID list). The combo of these two bugs added an extra copy of the intention's nodeID to the actor every time the intention was clicked. Now, these and related bugs are resolved, and the process is more efficient because intentions will only be removed and re-embeded when dragged by the user.

Linked issues: fixes #312

### DIFF
--- a/leaf-ui/js/loadSaveFunctions.js
+++ b/leaf-ui/js/loadSaveFunctions.js
@@ -59,7 +59,6 @@ reader.onload = function() {
 		        initSat = (initSat !== ' ' && initSat !== '')  ?  oldSatValToBinary[initSat] : '0000';
 			    var intentionEval = new UserEvaluation(intention.nodeID, '0', initSat);
 			    analysisRequest.userAssignmentsList.push(intentionEval);
-				console.log(analysisRequest.userAssignmentsList);
 		    
 		        // make the T upside down
 		        var satValText = cells[i].attrs['.satvalue'].text;
--- a/leaf-ui/js/loadSaveFunctions.js
+++ b/leaf-ui/js/loadSaveFunctions.js
@@ -51,7 +51,7 @@ reader.onload = function() {
 		    	cells[i]["nodeID"]  = intention.nodeID;
 
 		    	if (actorID !== '-') {
-		    		model.getActorByID(actorID).intentionIDs.push(intention.nodeID);
+		    		model.getActorByID(actorID).addIntentionID(intention.nodeID);
 		    	} 
 
 		        // create intention evaluation
@@ -59,6 +59,7 @@ reader.onload = function() {
 		        initSat = (initSat !== ' ' && initSat !== '')  ?  oldSatValToBinary[initSat] : '0000';
 			    var intentionEval = new UserEvaluation(intention.nodeID, '0', initSat);
 			    analysisRequest.userAssignmentsList.push(intentionEval);
+				console.log(analysisRequest.userAssignmentsList);
 		    
 		        // make the T upside down
 		        var satValText = cells[i].attrs['.satvalue'].text;
--- a/leaf-ui/js/object/Class.js
+++ b/leaf-ui/js/object/Class.js
@@ -1837,14 +1837,11 @@ class AnalysisRequest {
     }
 
     /**
-     * Removes all UserEvaluation objects in
-     * userAssignmentsList, with an intentionID equal to
-     * nodeID
+     * Removes the UserEvaluation object in
+     * userAssignmentsList that corresponds with nodeID
      *
      * @param {String}
      */
-
-
     removeIntention(nodeID) {
         for (var i = 0; i < this.userAssignmentsList.length; i++) {
             if (this.userAssignmentsList[i].intentionID == nodeID) {
--- a/leaf-ui/js/object/Class.js
+++ b/leaf-ui/js/object/Class.js
@@ -293,25 +293,29 @@ class Actor {
 
     /**
      * Removes intention ID nodeID from the
-     * intentionIDs array
+     * actor's intentionIDs array
      *
      * @param{String} nodeID
      */
-    removeIntentionID(nodeID,  userAssignmentsList) {
-        for (var i = 0; i < this.intentionIDs.length; i++) {
-            if (this.intentionIDs[i] == nodeID) {
-                this.intentionIDs.splice(i, 1);
-                return ;
-            }
-        }
-        while (i < userAssignmentsList.length) {
-            if (userAssignmentsList[i].intentionID == nodeID) {
-                userAssignmentsList.splice(i, 1);
-            } else {
-                i++;
-            }
-        }
+    removeIntentionID(nodeID) {
+        // remove node from intentionIDs list
+        var ind = this.intentionIDs.indexOf(nodeID);
+        this.intentionIDs.splice(ind, 1);
+    }
 
+    /**
+     * Adds intention ID nodeID to the
+     * intentionIDs array if not a duplicate
+     *
+     * @param{String} nodeID
+     */
+    addIntentionID(nodeID) {
+        // add if not in list already
+        if (!this.intentionIDs.includes(nodeID)) {
+            this.intentionIDs.push(nodeID);
+            // keep list sorted
+            this.intentionIDs.sort();
+        }
     }
 }
 Actor.numOfCreatedInstances = 0;
@@ -1842,13 +1846,10 @@ class AnalysisRequest {
 
 
     removeIntention(nodeID) {
-        var i = 0;
-
-        while (i < this.userAssignmentsList.length) {
+        for (var i = 0; i < this.userAssignmentsList.length; i++) {
             if (this.userAssignmentsList[i].intentionID == nodeID) {
                 this.userAssignmentsList.splice(i, 1);
-            } else {
-                i++;
+                break;
             }
         }
     }
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -1214,18 +1214,17 @@ graph.on('change:size', function(cell, size) {
 graph.on('remove', function(cell) {
     //TODO: What I have changed
     if(cell.isLink() && !(cell.prop("link-type") == 'NBT' || cell.prop("link-type") == 'NBD')){
-        //To remove link
+        // To remove link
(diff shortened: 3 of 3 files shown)


leaf-ui/js/loadSaveFunctions.js
```diff
@@ -51,7 +51,7 @@ reader.onload = function() {
 		    	cells[i]["nodeID"]  = intention.nodeID;
 
 		    	if (actorID !== '-') {
-		    		model.getActorByID(actorID).intentionIDs.push(intention.nodeID);
+		    		model.getActorByID(actorID).addIntentionID(intention.nodeID);
```
leshghi, line 54: Good fix

leaf-ui/js/object/Class.js
```diff
@@ -1842,13 +1846,10 @@ class AnalysisRequest {
```
leshghi: I know this is slightly out of scope of your changes but could you fix the extra lines between this method and its doc string?

leaf-ui/js/object/Class.js
```diff
@@ -1842,13 +1846,10 @@ class AnalysisRequest {
...
     removeIntention(nodeID) {
-        var i = 0;
-
-        while (i < this.userAssignmentsList.length) {
+        for (var i = 0; i < this.userAssignmentsList.length; i++) {
             if (this.userAssignmentsList[i].intentionID == nodeID) {
                 this.userAssignmentsList.splice(i, 1);
-            } else {
-                i++;
+                break;
```
leshghi, line 1849: Does the nodeID only match the intentionID in one element of the userAssignmentsList, hence only removing the first instance of a match? If so, you may also want to update the comment for this method to make that clear.

leaf-ui/js/onFunctions.js
```diff
@@ -1034,23 +1004,78 @@ paper.on('blank:pointerdown', function(evt, x, y) {
...
-		return;
-	}
-
-	var cell = cellView.model;
-	if (cell instanceof joint.dia.Link){
-		cell.reparent();
-	}
+paper.on({
+    'cell:pointerdown': function(cellView, evt, x, y) {
+        var interact = true;
```
leshghi, line 1010: Just so you know this interact variable doesn't actually appear to have any impact...I ran into this when working on #314 but didn't want to delete since I didn't need to touch it and wanted to test more extensively first but its possible you may be in the clear to delete.

leaf-ui/js/onFunctions.js
```diff
@@ -1034,23 +1004,78 @@ paper.on('blank:pointerdown', function(evt, x, y) {
...
-	if(mode == "Analysis"){
-		return;
-	}
-
-	var cell = cellView.model;
-	if (cell instanceof joint.dia.Link){
-		cell.reparent();
-	}
+paper.on({
+    'cell:pointerdown': function(cellView, evt, x, y) {
```
leshghi: For the first two cell: events cellView, x, and y are unused and for the last one x and y are unused.

leaf-ui/js/onFunctions.js
```diff
@@ -1034,23 +1004,78 @@ paper.on('blank:pointerdown', function(evt, x, y) {
...
+            var cell = cellView.model;
+            if (cell instanceof joint.dia.Link) {
+                if (evt.data.move){
+                    // if link moved, reparent
+                    cell.reparent();
+                    // check if link still valid
+                    basicActorLink(cell);
+                }
+            } else {
+                // Non-Link behavior
```
leshghi: Can this comment be deleted / or clarified w/r/t what code it is referring to

leaf-ui/js/onFunctions.js
```diff
@@ -1151,67 +1176,34 @@ function removeHighlight(){
...
-					var nodeID = cell.attributes.nodeID;
-					var actorID = actorCell.attributes.nodeID
-					model.getIntentionByID(nodeID).nodeActorID = actorID;
-					model.getActorByID(actorID).intentionIDs.push(nodeID);
-				}
-			}
-		}
-	}
-
+    var ActorsBelow = paper.findViewsFromPoint(cell.getBBox().center());
```
leshghi: I think it might be worth changing the name of this variable, since it is a bit misleading - paper.findViewsFromPoint() is returning all views at a certain point, this contains the cell itself and any overlapping elements (actors, goals, tasks, etc both above and below).

leaf-ui/js/onFunctions.js
```diff
@@ -1151,67 +1176,34 @@ function removeHighlight(){
...
+
+    // intention always returns with ActorsBelow
+    // so always at least one element in list
+    if (ActorsBelow.length > 1) {
```
leshghi: Optional: What if you took the return from paper.findViewsFromPoint() and just filtered it like paper.findViewsFromPoint().filter(view => view.model instanceof joint.shapes.basic.Actor))? Then you could check the list to see if it contained any elements and just iterate through those. And if not go into your else statement. I think it might be worth it just for readability but obviously up to you

leaf-ui/js/onFunctions.js
```diff
@@ -1230,18 +1222,17 @@ graph.on('change:size', function(cell, size) {
...
-        //To remove link
+        // To remove link
         var link = cell;
         clearInspector();
         model.removeLink(link.linkID);
     }
 
     else if((!cell.isLink()) && (!(cell["attributes"]["type"]=="basic.Actor"))){
-        //To remove intentions
+        // To remove intentions
```
leshghi, line 1237: lovelyyyy

leshghi (approved):
Kathleen this looks amazing to me thanks for taking all the extra effort to really go through rappid and not only fix this bug but also take the time to remove other stuff that wasn't working. Much appreciated! I did some testing and all looked good to me. I left a few minor comments mostly about cleanup and clarity so def address those if possible, but you have my full approval! Very nice great work and thanks again!

leaf-ui/js/object/Class.js
krhablutzel: Yes, there's one entry in the UAL for each nodeID; I'll note that.

leaf-ui/js/onFunctions.js
krhablutzel: It appears that the interact variable prevents the element inspector from appearing in the right panel, and also the menus that pop up to allow the user to delete nodes and add links from them. I tried setting` interact = true` for analysis mode and was able to delete nodes, and when I clicked on a node the element inspector replaced the analysis sidebar on the right.
Screen Shot 2021-04-25 at 17 35 14

leaf-ui/js/onFunctions.js
krhablutzel: I can remove the unused x and y, but removing cellView appears to cause errors because this function is passed evt as the second parameter.

leaf-ui/js/onFunctions.js
krhablutzel: Love this! Great idea!

leaf-ui/js/onFunctions.js
krhablutzel: Clarified that it's referring to the "else" statement!

leaf-ui/js/onFunctions.js
```diff
@@ -1034,23 +1003,83 @@ paper.on('blank:pointerdown', function(evt, x, y) {
...
+    'cell:pointermove': function(cellView, evt) {
+        if (!evt.data.move && evt.data.interact){
+            // start of a click and drag
+            evt.data.move = true;
+        }
+      },
+    'cell:pointerup': function(cellView, evt) {
+        // undefined event occurs due to paper.trigger("cell:pointerup"...
+        // on adding new cell to paper
+        if (evt == undefined) {
```
krhablutzel, line 1027: evt is undefined when we add a new cell to the paper, because we call paper.trigger("cell:pointerup" without passing in evt data from a pointerdown or pointermove. All the cell highlighting, inspector appearing, re-embedding  in actor, etc. that happens in pointerup should occur when a new cell is added, so this defines evt.data such that those will occur. (Adjusted this code since Ali's review earlier today—I was skipping undefined evts and calling an extra embedBasicActor when adding intentions, but it didn't preserve the highlighting and inspector behavior.)

leaf-ui/js/onFunctions.js
leshghi: Ooo very quirky not sure how I feel about that but I'm sure its something we can look into more during the refactor - thanks so much for double checking :)

PR #348, opened 2021-04-26
Model management "final" bug fixes

krhablutzel:
Working with Ali @leshghi on some final bugs in model-management before the launch of Version 1.5!

- Cleaned up the AnalysisConfiguration class structure
- Disabled ways to edit analysis configuration after generating results from it (won't fix intermediate values in this version)
- Fixed the renaming feature to remove bug and clean up CSS
- Fixed bug where analysisRequest didn't update after deleting a selected config
- Fixed sidebar loading bug where analysis config sidebar would briefly pop up on page load
- Removed individual event listeners from elements and put listeners in onFunctions per JavaScript convention
- Put css updates in their own functions in displayAnalysis
- Cleaned up some comments, empty spaces, and console logs

### DIFF
--- a/leaf-ui/js/displayAnalysis.js
+++ b/leaf-ui/js/displayAnalysis.js
@@ -200,9 +200,6 @@ function updateNodeValues(nodeID, satValue) {
 function addFirstAnalysisConfig(){
     // reset to default analysisRequest while preserving userAssignmentsList
     defaultAnalysisRequest();
-    // reset background colors
-    $(".config-elements").css("background-color", "");
-    $(".result-elements").css("background-color", "");
     // reset highest position to 1
     highestPosition = 1;
     // add config 1
@@ -312,9 +309,7 @@ function removeConfiguration(configElement) {
     // Highlight the currAnalysisConfig in the UI
     newConfigElement = document.getElementById(currAnalysisConfig.id);
     analysisRequest = currAnalysisConfig.getAnalysisRequest();
-    $(".config-elements").css("background-color", "");
-    $(".result-elements").css("background-color", "");
-    $(newConfigElement.querySelector('.config-elements')).css("background-color", "#A9A9A9");
+    switchSelectedShadingConfig(newConfigElement);
     // Remove full configuration div (includes results)
     configElement.remove();
     // Remove config from analysisMap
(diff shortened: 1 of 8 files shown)


esterzhao (changes requested):
Did some pretty extensive use testing, found that issue #320 is not resolved on this branch or on model-managment, which may have been a casualty of merge conflict resolution. 

Other than that, everything is working well and these are some great cleanups, thanks guys!

leshghi:
> ok awesome! just wanted to make sure it wasn't overlooked

Haha me too it took me a second to remember why I was clear to remove it I had to double check

leaf-ui/js/displayAnalysis.js
```diff
@@ -185,27 +185,32 @@ function updateNodeValues(nodeID, satValue) {
...
  */
  function loadAnalysisConfig(){
-    // Refresh the analysis sidebar to reflect current analysis request values
-    refreshAnalysisUI();
     // if there are no configs in the map
     if(analysisMap.size == 0){
         // Add a new, empty config to the map
         addFirstAnalysisConfig();
     }
+    refreshAnalysisUI();
```
leshghi, line 194: Since refreshAnalysisUI() is now also called in addFirstAnalysisConfig() maybe we only need to call it if we are not calling addFirstAnalysisConfig()?

leaf-ui/js/displayAnalysis.js
```diff
@@ -554,3 +571,16 @@ function getConfigHtml(id){
             '<div class="dropdown-container"></div>' +
            '</div>';
 }
+
+/**
+ * Reset global analysisRequest to default analysisRequest settings
+ * while preserving userAssignmentsList
+ */
+function defaultAnalysisRequest(){
```
leshghi: super optional: consider renaming to something like resetToDefault etc etc because calling defaultAnalysisRequest makes me feel like it should be returning something instead of updating a global variable...

leshghi (approved):
Looks good to me, @krhablutzel I looked at your commit changes again and made a few super minor comments. That said I know we did most of these changes together so I don't think my approval carries much weight.

leaf-ui/js/displayAnalysis.js
krhablutzel: I can add an else!

esterzhao (approved):
:)

PR #351, opened 2021-04-30
Create Backbone models for analysis mode

leshghi:
Extremely basic skeleton models to get started on implementing backbone structure for analysis mode.

### DIFF
--- a/leaf-ui/rappid-extensions/AnalysisModels.js
+++ b/leaf-ui/rappid-extensions/AnalysisModels.js
@@ -11,11 +11,12 @@
 // TODO: Break AnalysisResult into params
 var ResultModel = Backbone.Model.extend({
     idAttribute: "uid",
+
     defaults: {
         name:"Default Result",
         analysisResult: new AnalysisResult(),
-    }
-})
+    },
+});
 
 /**
  * Collection that holds results for analysis configurations
@@ -24,8 +25,10 @@ var ResultModel = Backbone.Model.extend({
  */
 var ResultCollection = Backbone.Collection.extend({
     model: ResultModel,
+    
+    /** Used for eventually storing and accessing collection on server */
     url: "/results",
-})
+});
 
 /**
  * Model for analysis configurations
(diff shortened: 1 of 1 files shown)


leaf-ui/rappid-extensions/AnalysisModels.js
```diff
@@ -0,0 +1,88 @@
...
+    model: ConfigModel,
+    url: "/configurations",
+
+    initialize: function(){
+        this.on('add', this.onSelectedChanged, this);
+    },
+
+    // When config is clicked/selected, find previously selected config and unselect it. 
+    onSelectedChanged : function(changedConfig){
+        this.filter(configModel => configModel.get('selected') == true && changedConfig != configModel)
```
ibrodycalixto, line 85: These arrow functions r so fancy...

leaf-ui/rappid-extensions/AnalysisModels.js
```diff
@@ -0,0 +1,88 @@
...
+})
+
+/**
+ * Collection that holds results for analysis configurations
+ * 
+ * {ResultModel} model
+ */
+var ResultCollection = Backbone.Collection.extend({
+    model: ResultModel,
+    url: "/results",
```
ibrodycalixto, line 30: Is the idea with collection urls to store all the results/configs in a new place on the server?

ibrodycalixto (approved):
From my knowledge of backbone (lol) looks like a good start to me. I think it works great as an example of how models/collections work.

leaf-ui/rappid-extensions/AnalysisModels.js
leshghi: Sort of, but to be honest I don't think it is necessary right now because that is not currently really how the communication between frontend/backend is set up. Happy to remove for now!

ibrodycalixto: Maybe just comment to clarify!

Issue #329, opened 2021-04-10
LinkInspector does not remove "NB" from source and target after selecting "Not Both" options for link


leshghi:

leshghi:
Check out line 171 to see some code that might be useful!

PR #360, opened 2021-05-03
Delete contraints inspector file

leshghi:
File is never accessed, can be deleted. Was replaced by LinkInspector.


ibrodycalixto (approved):
Looks good! Found a few bugs, but nothing that isn't on develop.

leshghi:
> Looks good! Found a few bugs, but nothing that isn't on develop.

Ah yeah I just saw you added #364 which I think is the same as #329. Did you find any other bugs? We could reference them in the LinkInspector.js walkthrough issue

Issue #365, opened 2021-05-04
Rename click outside box event doesn't occur when clicking on paper/graph


leshghi:
Current only an issue on config-sidebar branch, but may persist so want to document now.

YesugenB:
This is an issue in the backbone-refactor branch.

Issue #353, opened 2021-04-30
Revamp Analysis mode right sidebar (in AnalysisInspector) to take in a model and correctly implement backbone


leshghi:
Also includes moving IVT into it's own file for now.

PR #366, opened 2021-05-05
Config sidebar

### LINKED ISSUE #352: Convert Config sidebar to backbone structure

leshghi:
This is currently just a UI representation in backbone of the config sidebar. The functionality has been broken, and will not fully be implemented until #353 is initially complete, since we need those two backbone files to talk to each other. I thought it was best to get what we do have in now so that #353 can access this UI structure and we can begin to connect the two moving forward. The most glaring UI bug associated with this branch is #365, but of course since functionality is broken there is a myriad of other bugs. Once we get #353 set up we can start to name and create issues for each individual one. This pull request will also comment out most of the previous config sidebar code (main in displayAnalysis.js). I am holding off on deleting it until we have fully recreated functionality, because it will be useful as reference until then.


leaf-ui/js/displayAnalysis.js
```diff
@@ -198,121 +198,123 @@ function updateNodeValues(nodeID, satValue) {
...
-    // Set default UAL to preserve in future configs
-    // It is necessary to push each UAL seperately 
-    // to avoid the defaultUAL variable updating along with currAnalysisConfig
-    currAnalysisConfig.userAssignmentsList.forEach(uAL => defaultUAL.push(uAL));
-    analysisRequest = currAnalysisConfig.getAnalysisRequest();
-    switchConfigs(firstConfigElement);
-    // Refresh the sidebar to include the config vars
-    refreshAnalysisUI();
-}
+// function loadAnalysis(){
```
ibrodycalixto, line 221: It seems like we need to make it so that loadSave functions populates the config/results collections...would that be part of the functionality you want to implement here or later?

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -0,0 +1,308 @@
...
+        '</div>',
+       '</div>',
+       '</script>'].join(''),
+
+    events: {
+        'click .config-elements': 'switchConfig',
+        'click .deleteconfig-button': 'removeConfig',
+        'click .dropdown-button' : 'toggleDropdown',
+        'dblclick .config-elements': 'rename',
+        'blur .config-input': 'setConfigName',
```
ibrodycalixto, line 130: Does this listener check if user is still typing?

ibrodycalixto (commented):
I'm not sure I know enough about views to correct mistakes if there are any, but from what I understand all the functionality and organization works for me. I'll get started on the left side bar model and view tonight so we can restore functionality to analysis mode

leaf-ui/js/displayAnalysis.js
leshghi: Yeah, a ton of functionality like load/save is currently broken. This will most likely be one of the last parts we touch, as it will involve reworking some of the conversions to JSON (Since the structure of our data is changing). I think for now the first step is to get all our UI pieces in the right format, and then we can start touching things that talk to the backend.

leaf-ui/rappid-extensions/ConfigInspector.js
leshghi: Are you referring to the .keyup event? It is triggered on any keyup event while the user is typing in the config-input, but the function itself is only used if the user presses the enter key to "escape" editing and save the new name.

leshghi: Should I add a more clarifying comment?

ibrodycalixto: I think it's fine!

PR #367, opened 2021-05-07
Dead code cleanup

### LINKED ISSUE #356: Remove the code to load in files from the old version of BloomingLeaf from loadSaveFunctions.js
Important to clearly mark this change, as we need to document when we removed this ability.

### LINKED ISSUE #358: Remove accessDatabase code from onFunctions.js

leshghi:
- Removes legacy DataBase functions (accessDatabase, accessDatabaseWithRead, updateDataBase). Closes #358 
- Removes ability to load in JSON files from old version of BloomingLeaf. Closes #356 
- Removes tracking ability, as it is not fully implemented and the use case for tracking is not clear for the current version and usage of the tool.

I tested to best of my abilities and everything seems to still be functioning, but would appreciate a quick lookover from @amgrubb since they know this legacy code and it's purpose.


amgrubb (approved):
Great work. Remove pop-up.

PR #369, opened 2021-05-16
Class migration

### LINKED ISSUE #339: Pre-backbone refactor architecture changes
Make proposed changes from https://docs.google.com/document/d/1FsTrmmP5HoIy2ko9rFivR28-arcb7VfJ5JmRN9DFOlU/edit?ts=607627dd#heading=h.vkevl31ilfv0

Primarily class migration

krhablutzel:
Implements the proposed architecture changes as described in the [Frontend Refactor Proposal](https://docs.google.com/document/d/1FsTrmmP5HoIy2ko9rFivR28-arcb7VfJ5JmRN9DFOlU) by migrating classes from `js/object/Class.js` into `object/analysisObjects.js`, `object/modelObjects.js`, and `object/analysisView.js`.


leshghi (approved):
Okay, I took a look at this and it seems good to me. Since this pull request is basically all just moving classes around with no change to logic (with the exception of the SliderObj class), we can just approve it as a general "moving of code" request. Originally when I reviewed I was commenting improvements to classes/class methods, but I think that is in some ways out of scope. Even though the code is highlighted as a change, this is just because of the migrating into new files and not because the request is actually touching the code/logic itself. So I am going to approve this under the condition that before it is merged issues are created to review and cleanup each file. This cleanup can be done later/as we move forward with the refactor but I want to make sure it is documented as a "Todo" in the meantime.

PR #382, opened 2021-06-08
removed 'onclick' functions from analysis.html

### LINKED ISSUE #355: Remove all "onclick" functions from analysis.html, replace with event listeners
Same as #354 but for analysis.html (the Explore next States window). I would suggest migrating the code in analysis.js to the currently empty onFunctionsNextState.js file as well as part of this issue.

YesugenB:
removed 'onclick' functions from analysis.html, replaced with event listeners. analysis.js contains functions, onFunctionsNextState contains event listeners

### DIFF
--- a/leaf-ui/analysis.html
+++ b/leaf-ui/analysis.html
@@ -40,17 +40,17 @@
                 <button id="btn-save" class="btn">Save</button>
             </div> -->
 
-            <button id="btn-zoom-in" class="btn" onclick="zoomIn(analysis.paperScroller)">Zoom In</button>
-            <button id="btn-zoom-out" class="btn" onclick="zoomOut(analysis.paperScroller)">Zoom Out</button>
+            <button id="btn-zoom-in" class="btn">Zoom In</button>
+            <button id="btn-zoom-out" class="btn">Zoom Out</button>
 
             <!-- <button id="btn-save-leaf" class="btn"> Export .leaf </button> -->
 
             <div class="dropdown">
               <div class="btn">Font Size</div>
               <div class="dropdown-toolbar">
-                <a id="btn-fnt" onclick="defaultFont(analysis.paper)">Default</a>
-                <a id="btn-fnt-up" onclick="fontUp(analysis.paper)">Increase</a>
-                <a id="btn-fnt-down" onclick="fontDown(analysis.paper)">Decrease</a>
+                <a id="btn-fnt">Default</a>
+                <a id="btn-fnt-up">Increase</a>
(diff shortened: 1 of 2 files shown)


leaf-ui/js/onFunctionsNextState.js
```diff
@@ -7,4 +7,26 @@
...
+// }
+
+//Only event listeners
+$('#btn-zoom-in').on('click', function(){ zoomIn(analysis.paperScroller); });
+$('#btn-zoom-out').on('click', function(){ zoomOut(analysis.paperScroller); });
+$('#btn-fnt').on('click', function(){ defaultFont(analysis.paper);});
+$('#btn-fnt-up').on('click', function(){  fontUp(analysis.paper);});
+$('#btn-fnt-down').on('click', function(){ fontDown(analysis.paper);});
+$('#inspector-btn-small').on('click', function(){ goToState(); });
+$('#conflictFl').on('click', function(){ add_filter();});
```
leshghi: For all the add_filter() params, since they all call the same function regardless of id, you can probably just use the class name to get it all in one line. Aka $('.filter_checkbox').on('click', function(){add_filter});

leaf-ui/analysis.html
```diff
@@ -89,23 +89,25 @@
...
             <br>
             <label>Actions</label>
-            <button class="inspector-btn" onclick="save_current_state()">Save &amp Close</button>
-            <button class="inspector-btn" onclick="generate_next_states()">Explore Next States</button>
-            <button class="inspector-btn" onclick="window.close();">Close</button>
+            <button id="saveClose" class="inspector-btn">Save &amp Close</button> 
+            <button id="exploreNextSates" class="inspector-btn">Explore Next States</button>
+            <button id="close" class="inspector-btn">Close</button>
+
+
```
leshghi: Remove extra spaces

leshghi (approved):
Looks great to me - just make the one change I suggested and you should be all set!

leaf-ui/js/onFunctionsNextState.js
YesugenB: $('.filter_checkbox').on('click', function(){add_filter}); wasn't working for some reason? You could still select filters but they weren't actually filtering. Not sure why. Should I look more into this?

leshghi: I forgot to put the () after add_filter - could that be it? Either way I would probably look into it more if you don't mind cause it could save a lot of lines of code and simplify stuff down

PR #384, opened 2021-06-09
Issue376 Fixed in LinkInspector

### LINKED ISSUE #376: In LinkInspector.js, evolving relationships that start with and/or should be able to end with and/or

Manarnz:
I fixed the bug that does not allow users to choose an end relationship when they select and/or as the begin.

### DIFF
--- a/leaf-ui/rappid-extensions/LinkInspector.js
+++ b/leaf-ui/rappid-extensions/LinkInspector.js
@@ -63,7 +63,6 @@ var LinkInspector = Backbone.View.extend({
     render: function(cell) {
         this.cell = cell;
         this.link = model.getLinkByID(cell.attributes.linkID);
-
         // Selecting which template to render ACTOR-LINK or INTENTIONS-LINK
         if (cell.prop('linktype')) {
             this.$el.html(_.template(this.actortemplate)());
@@ -106,7 +105,6 @@ var LinkInspector = Backbone.View.extend({
     switchMode: function() {
 
         var type = this.cell.attributes.labels[0].attrs.text.text;
-
         // array of link values
         var values = type.split("|");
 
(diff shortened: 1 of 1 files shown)


leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -248,8 +253,7 @@ var LinkInspector = Backbone.View.extend({
             this.link.linkType = begin.toUpperCase();
             this.link.postType = 'NO';
 
-            $("#repeat-error").text("Saved!");
-            $("#repeat-error").css("color", "lightgreen");
+
```
leshghi, line 248: Remove extra line

leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -170,11 +170,16 @@ var LinkInspector = Backbone.View.extend({
             
         } else {
             //Verify if it is possible to remove the NB tag from source and target
-            if (!this.hasNBLink(source, this.cell) && this.hasNBTag(source)){
-                source.attr(".funcvalue/text", "");
+            if (!this.hasNBLink(source, this.cell) && !this.hasNBTag(source)){
```
leshghi: Lets talk about how to make sure these changes don't show up in your pull request since they should already be merged into backbone-refactor

leshghi (changes requested):
This looks great to me - good fix! There is a new issue I found after your fix, which is that if you click away from the LinkInspector and then click back in, the link resets to a "null" value. However, I you did not create that bug, it was just uncovered after you fixed this one! I would recommend checking out the bug I just mentioned, and adding a descriptive Issue for it in our project board with screenshots if possible (I can show you how to do this if you want help). That way it will be documented for a future fix :) Awesome job again - just make the changes I mentioned and the Issue on github and then I'll re-review and you should be all set.

leshghi (approved):
Looks great! Make sure to chose "squash and merge" from the green merge button dropdown menu and then you should be all set!

PR #388, opened 2021-06-10
Issue #372 ElementInspector.js no-value dropdown fix and Issue #377 modelObjects.js UD function drop down fix

### LINKED ISSUE #372: In ElementInspector.js user defined doesn't work if initial value is no value
bugScreenshot1

### LINKED ISSUE #377: In ElementInspector.js, UD functions should be able to select a value for the first dropdown
The correct dropdown options should render for the first value. See below for current state: Screen Shot 2021-06-08 at 12 55 57 PM

i-m-foster:
Closes #372 and closes #377


leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -231,9 +231,7 @@ var ElementInspector = Backbone.View.extend({
         if (this.intention.getInitialSatValue() == '(no value)') {
             // remove current options, add 3 options
```
leshghi: Update this comment to explain that stochastic should be only option

leshghi (changes requested):
This looks great - just remember to change the default value so that it doesn't switch from "no value" to None then you're all set

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -14,7 +14,8 @@ var validPair = {
     },
     "R":{
         "validInitValue": ["none", "satisfied", "partiallysatisfied", "denied", "partiallydenied", "(no value)"],
-        "defaultValue": ["none"]
+        // default value changed to (no value)
```
leshghi: Just capitalize default and you are all set!

leshghi (approved):
Awesome job!!!

PR #392, opened 2021-06-10
modified actor model in joint.extentions

Domily11:
Added actortype parameter in joint.shapes.basic.Actor and placed a function to assign associated lines. Wouldn't show up without updating actorinspector


Manarnz (approved):
Good job! These changes are good, let's keep working on bugs

PR #397, opened 2021-06-11
Analysis inspector backbone

### LINKED ISSUE #353: Revamp Analysis mode right sidebar (in AnalysisInspector) to take in a model and correctly implement backbone

YesugenB:
First step towards migrating analysis inspector to a more complete backbone structure.
Includes:
- Addition of ConfigModel passed into backbone view
- Migration of IVT to another file
- Use of script template

Outstanding Issues:
- When removing a config, we need to clear inspector and reselect new configuration and/or result #407
- IVT table needs to be properly reformatted into a backbone view with a model #406 
- Communication with backend is broken (this is expected behavior for this point) and we will need to determine how to best package and send the backbone models to the backend as we continue this refactor

I am sure there are many other small bugs, but this should be a good first step.

### DIFF
--- a/leaf-ui/css/analysis_style.css
+++ b/leaf-ui/css/analysis_style.css
@@ -51,7 +51,6 @@
 
 /*Hide the intermTable dialog by default*/
 .intermT {
-    display: none;  /*Hidden by default */
     position: fixed; /* Stay in place */
     z-index: 1; /* Sit on top */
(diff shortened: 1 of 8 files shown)


YesugenB (commented):
I think it looks good!

amgrubb (approved):
Great work. Good comments.

PR #399, opened 2021-06-11
Removed "onclick" functions from index.html

### LINKED ISSUE #354: Remove all "onclick" functions from index.html, replace with event listeners in onFunctions
For example, in index.html, the zoom in button looks like this:
`<button id="btn-zoom-in" class="btn" onclick = "zoomIn(paperScroller)">Zoom In</button>`
Instead of having onclick="zoomIn(paperScroller)", which calls a function zoomIn in onFunctionsBothWindows.js, we should have an event listener that listens to the click event like:
`$('#btn-zoom-in').on('click', function(){
          zoomIn(paperScroller);
};`
Although this may look more complex/verbose, it is [actually correct practice](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#inline_event_handlers_%E2%80%94_dont_use_these).

YesugenB:
Replaced with event listeners in onFunctions.js

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -4,7 +4,7 @@ It also contains the setup for Rappid elements.
 */
 
 /**
- * Event listeners for index.html
+ * Event listeners for index.html toolbar functions
  */
  $('#btn-zoom-in').on('click', function(){ zoomIn(paperScroller); });
  $('#btn-zoom-out').on('click', function(){ zoomOut(paperScroller); });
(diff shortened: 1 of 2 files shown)


leaf-ui/js/onFunctions.js
```diff
@@ -3,6 +3,17 @@ This file contains all the jQuery functions that are associated with buttons and
 It also contains the setup for Rappid elements.
 */
 
+/**
+ * Event listeners for index.html
```
leshghi:
```suggestion
 * Event listeners for index.html toolbar functions
```

leshghi (approved):
This looks great to me! Typically I would suggest leaving comments but in this case I feel that they are pretty self explanatory...Nice work - thank you :)

PR #400, opened 2021-06-11
Fixed issue #389 

### LINKED ISSUE #389: In LinkInspector.js when switching back to evolving with begin as and/or end link becomes null
Screenshot 2021-06-10 132757

When selecting AND/OR as begin, and choosing an end from template A then switching back, the end link shown in the graph becomes null (in the bar, it stays the same).

Manarnz:
Before, when the begin and end are AND/OR, switching back to Evolving Relationship would make the end "null." I fixed this issue.


leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -228,14 +227,10 @@ var LinkInspector = Backbone.View.extend({
...
         } else if(begin == "and" || begin == "or") {
             this.setSelectValues('#link-type-end', 'A');
-
-            var end = $("#link-type-end").val();
-
             $("#link-type-end").prop('disabled', false);
             $("#link-type-end").css("background-color","");
             
             //Saving this option
-            this.cell.prop("link-type", begin + "|" + end);
```
leshghi: Why did you delete this line? Asking just because I don't know as much about link-type properties as you do, and just want to make sure we are still saving the type of link into the cell somewhere

leshghi (commented):
Looks good to me - just answer my question below and once I understand why that line was deleted I'll approve so you can merge

leaf-ui/rappid-extensions/LinkInspector.js
Manarnz: When I commented it out, it didn't make a difference because we have another line after it that does the same thing, so I deleted it because it is repetitive.

leshghi: I looked but couldn't find the line you are referring to that does the same thing - could you direct me to it (just send me the number of the line in your branch)? Thanks!

leshghi (commented):
Also - did you mean to make this pull request for a merge into Issue374?

leaf-ui/rappid-extensions/LinkInspector.js
Manarnz: line 261, in updateEndEvolRelations. It is also important to note that the command I deleted was only in the AND/OR if statement, the other if statements did not include it, even the ones that have a begin and an end.

Manarnz:
> Also - did you mean to make this pull request for a merge into Issue374?

Yes, because it depends on the changes I made in issue374. is that okay?

leaf-ui/rappid-extensions/LinkInspector.js
leshghi: I see that, but as far as I can tell updateEndEvolRelations is not always called after updateBeginEvolRelations. For example, if someone changes the beginEvolRelationship, now that will not update the link-type property. Whether or not we need it to I think is a question we can work through as linkInspector.js is implemented with backbone, but in the meantime I think deleting this line is out of scope of the pull request so I would add it back in temporarily until we can look into it more

leshghi:
> > Also - did you mean to make this pull request for a merge into Issue374?
> 
> Yes, because it depends on the changes I made in issue374. is that okay?

I don't think it should be dependent on your Issue374 branch, what part specifically is it dependent on?

Manarnz:
> > > Also - did you mean to make this pull request for a merge into Issue374?
> > 
> > 
> > Yes, because it depends on the changes I made in issue374. is that okay?
> 
> I don't think it should be dependent on your Issue374 branch, what part specifically is it dependent on?

Okay done!

leshghi (changes requested):
When you click back into the linkInspector.js - although the text on the graph stays correct, the "end" dropdown menu gets reset to just say "end." We want it to say the same value as the link, so once you do that fix you should be set.

leshghi (approved):
Looks great! Congrats on finishing this one up :)

PR #402, opened 2021-06-14
began to implement backbone model in view in ElementInspector.js

i-m-foster:
would appreciate feedback on our progress


leaf-ui/js/onFunctions.js
```diff
@@ -942,7 +942,12 @@ paper.on({
                 if (cell instanceof joint.shapes.basic.Actor) {
                     actorInspector.render(cell);
                 } else {
-                    elementInspector.render(cell);
+                    var elementInspector = new ElementInspector({model: cell});
+                    console.log(elementInspector)
+                    // HTML moved here from initializeElements
+                    $('.inspector').append(elementInspector.el);
+                    elementInspector.switchModel(cell)
```
leshghi, line 910: You shouldn't need a switchModel method here since you already passed the cell in as the model in line 945

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -171,13 +179,14 @@ var ElementInspector = Backbone.View.extend({
...
+    
+    render: function() {
 
         // Save the Intention object from the global model variable to
-        // this.intention
-        this.intention = model.getIntentionByID(this.cell.attributes.nodeID);
+        //this.model.intention
+        console.log(this.model);
+        console.log(this.model.intention);
+        this.model.intention = model.getIntentionByID(this.model.attributes.nodeID);
```
leshghi: Remove this line because currently it is altering the structure/values of the cell associated with this view (The cell is this.model inside this view). If you want to store the intentions for now you could store them inside a parameter named this.intentions.

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -171,13 +179,14 @@ var ElementInspector = Backbone.View.extend({
...
 
         this.$el.html(_.template(this.template)());
```
leshghi:
```suggestion
        this.$el.html(_.template($(this.template).html())(this.model.toJSON()));
```
This should inject your model parameters directly into the template

leshghi (changes requested):
Great job injecting the cell model into the view! Some thoughts for next steps:
1. Use my suggestion for line 191 to inject the model parameters into the template. You will also need to turn the html template into a script. This can be done by using the tags: '<script type="text/template" id="item-template">',  '</script>' at the beginning and end of the template respectively. From there, you can do things like set text/values/etc directly inside the template. See ConfigInspector.js lines 109-123 for an example. You may need to alter the Intention model in joint.extensions.js to make sure all your parameters are accessible.
2. Once you have done that I would focus on making sure you clear the inspector view before adding a new one every time a cell is clicked. I can help you with that but I also have a [commit with a solution](https://github.com/amgrubb/BloomingLeaf/commit/b5da54dea4547937d216aaab0d4e82cbf61101c1) that you can look at and try to extrapolate from. 
3. The ElementInspector.js file has the most moving parts w/r/t parameters/models/logic so from there I might have to chat with @amgrubb  to find the best way to structure it all so that you can move away from using the global model object. 

Let me know if you get stuck/confused and I will help out! Thanks for sharing your progress :)

PR #412, opened 2021-06-15
Moved onclick event to onFunctions.js

### LINKED ISSUE #359: Go through keyboardFunctions.js
Determine if the functions work, how they work, and if they need their own separate file.
Move the window.onclick event to onFunctions.js

YesugenB:
Copy paste functions work


leshghi (commented):
This looks good to me - but thoughts on moving the keyboard functions to the onFunctionsBothWindows file? Just to consolidate non tool-specific related functions.

YesugenB:
Do you mean moving all of keyboardFunctions.js to onFunctionsBothWindows and then deleting the empty file?

leshghi:
> Do you mean moving all of keyboardFunctions.js to onFunctionsBothWindows and then deleting the empty file?

Yup!

leshghi (approved):
Looks good to me!

PR #422, opened 2021-06-16
Cleanup displayAnalysis

YesugenB:


leaf-ui/rappid-extensions/BackboneModels.js
```diff
@@ -1 +1,27 @@
-/** This file contains backbone model representations of the original model objects - WIP */
\ No newline at end of file
+/** This file contains backbone model representations of the original model objects - WIP */
```
leshghi: Remove this file from commit

leaf-ui/js/displayAnalysis.js
```diff
@@ -177,144 +177,6 @@ function updateNodeValues(nodeID, satValue) {
     }
 }
 
-/**
```
leshghi, line 180: I think you should also be able to delete a lot of the configuration sidebar specific functions as well (for example: switchConfigs, switchSelectedShadingConfig, etc).

leshghi (changes requested):
This looks good and I honestly think we could probably remove even more. For example the clear results/clear config functions don't really work anymore at the moment. We could rewrite them to clear from the configCollection instead of the analysisMap (since we are trying to phase out analysisMap). I don't know if that is worth doing yet though as the hope will be we can figure those out so that the configCollection does not have to be a global variable. Anyway - up to you what you decide to do, but we can also always just delete those and then make an issue to re-add them back in.

leaf-ui/js/displayAnalysis.js
```diff
@@ -574,18 +236,6 @@ function resetToDefault(){
...
- * @param {JQueryElement} configElement 
- */
-function switchSelectedShadingConfig(configElement){
-    $(".result-elements").css("background-color", "");
-    $(".config-elements").css("background-color", "");
-    $(configElement.querySelector(".config-elements")).css("background-color","#A9A9A9");
-}
-
 /**
  * Updates shading so that correct configuration and result is highlighted
```
leshghi: You can delete this too

leaf-ui/js/displayAnalysis.js
```diff
@@ -510,7 +211,7 @@ function switchResults(resultElement, configElement){
     analysisRequest = currAnalysisConfig.getAnalysisRequest();
 
     // Update UI accordingly
-    switchSelectedShadingResult(resultElement, configElement);
+    ShadingResult(resultElement, configElement);
```
leshghi: You can delete this

leaf-ui/js/onFunctions.js
```diff
@@ -693,18 +640,13 @@ $('#btn-clear-cycle').on('click',function(){
...
     // remove all configs from analysisMap
     analysisMap.clear();
-	// add back first default analysis config
-    addFirstAnalysisConfig();
     // reset graph to initial values
     revertNodeValuesToInitial();
 });
 
 $('#btn-clear-results').on('click', function() {
-    clearResults();
```
leshghi, line 707: Maybe remove refreshAnalysisUI() and leave a TODO in this clear result/clear config functions as well to go back and re-implement

leshghi (changes requested):
almost done!

leshghi (approved):
Looks great! I added one commit with some TODOs and you're all set!

PR #428, opened 2021-06-17
Removed CGI, Cycle Detection Call, and Input Files

amgrubb:
Remove CGI and files for cycle detection.
Removing cycle detection from analysis button.
Removed InputX files. This are only used in the cycle detection code. This will need to be updated after the backbone refactoring is complete.


leaf-ui/js/initializeElements.js
```diff
@@ -75,7 +75,7 @@ var mode;
 var showEditingWarning = true;
 
 var model = new Model();
-var analysisRequest = new AnalysisRequest();
+var analysisRequest = new AnalysisRequest();        //TODO: make not global!
```
leshghi, line 78: TODO: Remove (?)

leshghi (approved):
Looks good to me, feels good to delete things!

PR #437, opened 2021-06-22
created EvolvingFunctionBBM and IntentionBBM 

i-m-foster:
We would appreciate guidance ty


leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+    },
+
+    /**
+     * Returns the 4 digit representation for this
+     * EvolvingFunction's second last function segment's
+     * marked value value. If there is no second last function
+     * segment, this function returns the 4 digit representation of 
+     * the only function segment's marked value
+     */
+    getSecondLastMarkedVal: function() {
```
leshghi: Does it make sense to combine this and getLastMarkedVal and just pass in how far back you want to go as an integer parameter?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     *
+     * @param {String} time1
+     *   first relative time point
+     * @param {String} time2
+     *   second relative time point
+     */
+    // TODO fix this function
+    // I think we should set it so that the repeating function starts at time 1 and ends at 
+    // time 2
+    setRepeatingFunction: function(time1, time2) {
```
leshghi: Isn't repAbsTime an input box in the frontend? Can you get the repAbsTime from there? Or does it make more sense to remove the input box if repAbsTime is just predetermined from time1 and time2? Also - is repAbsTime used for anything in this function? ALSO is time1 and time2 are strings how are we subtracting them? Lots of questions sorry!

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     *
+     * Precondition: This EvolvingFunction must contain a RepFuncSegment
+     *
+     * @returns {String}
+     *   ex: 'A'
+     */
+    // TODO - fix this function 
+    getStartRepeatEpoch: function() {
+        if (this.repStart != 0) {
+            var repStartIndex = this.repStart.charCodeAt() - 64; // this should be the index of the first repeating segment
```
leshghi: Can we get a comment explaining why this is the index of the first repeating segment

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     * ends
+     *
+     * Precondition: This EvolvingFunction must contain a RepFuncSegment
+     *
+     * @returns {String}
+     *   ex: 'C'
+     */
+    // TODO - fix this function
+     getEndRepeatEpoch: function() {
+        var repStopIndex = this.repStop.charCodeAt() - 64; // this should be the index of the last repeating segment
```
leshghi: Add explanation of why here as well

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     * getMarkedVal: function(i) {},  
+     * findSegmentByStartTime: function(time) {}, 
+     */
+
+
 });
 
 var IntentionBBM = Backbone.Model.extend({
     initialize: function(options) { 
         this.nodeType = options.nodeType;
```
leshghi: This will prevent probably a lot of future errors - see message in slack surf21 channel
```suggestion
        _.extend({}, this.defaults, options)
```

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     */
+
+
 });
 
 var IntentionBBM = Backbone.Model.extend({
     initialize: function(options) { 
         this.nodeType = options.nodeType;  
     }, 
     defaults: {
```
leshghi, line 150:
```suggestion
    defaults: { 
         nodeType: null
```

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+    // TODO - update this function
+    // if we are using the automatically generated 'id', then we can just call 
+    // it whenever we need?? and would we be not allowed to change it??
+    setNewID: function(newID) {
+        var oldID = this.get('id');   
+        this.set('id', newID); 
+        this.evolvingFunction.set('id', newID); 
+ 
+        // UserAssignementsList
+        var userAssignList = analysisRequest.userAssignmentsList;
```
leshghi: We don't want to access the analysisRequest global variable since we are trying to phase it out. In the future the userAssignmentsList will be stored on the graph object.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        this.changeInitialSatValue('(no value)');
+    },    
+
+    /**
+     * Sets newID as the new ID for this intention
+     */
+    // TODO - update this function
+    // if we are using the automatically generated 'id', then we can just call 
+    // it whenever we need?? and would we be not allowed to change it??
+    setNewID: function(newID) {
```
leshghi: Can delete this whole function - ids will get automatically assigned and should not be reassigned

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+ 
+    }, 
+ 
+    /**
+     * Returns the 4 digit representation for this
+     * Intention's initial satisfaction value
+     *
+     * @returns {String}
+     */
+    getInitialSatValue: function() {
```
leshghi, line 209: Remove this for now since we don't want to access the analysisRequest - lets chat with @amgrubb about how to work this

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+            return intentionEval.evaluationValue;
+        }
+    }, 
+ 
+    /**
+     * Resets dynamic function
+     */
+    removeFunction: function() {
+        this.removeAbsCosnt();
+        this.evolvingFunction = new EvolvingFunctionBBM();
```
leshghi: If we are removing the evolving function should we just set it to null?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        } else {
+            return intentionEval.evaluationValue;
+        }
+    }, 
+ 
+    /**
+     * Resets dynamic function
+     */
+    removeFunction: function() {
+        this.removeAbsCosnt();
```
leshghi: Does cosnt mean constraint? It looks like const(ant) spelled incorrectly so maybe we could rename?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     * representing an absolute constraint, if requried.
+     *
+     * If funcType is RC, CR, MP, MN, SD, DS  a Constraint object
+     * representing an absolute constraint will be added. If not, this
+     * function does not do anything
+     *
+     * @param {String} funcType
+     *   ex: 'RC'
+     */
+    addAbsConst: function(funcType) {
```
leshghi: Removing constraints list, lets talk with @amgrubb about how to rework this

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+ 
+ 
+    /**
+     * Returns the absolute time for this Intention's absolute constraint at
+     * the starting epoch boundary start
+     *
+     * @param {String} source
+     *  ex. 'A'
+     */
+    getAbsConstTime: function(source) {
```
leshghi: I think this will actually be accessing the absTP param on the relevant function segment now. @amgrubb is that correct?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        return model.getAbsConstBySrcID(this.get('id'), source).absoluteValue;
+    }, 
+ 
+    /**
+     * Removes the absolute Constraint object(s) for this Intention from
+     * the global model variable, if such absolute Constraint object(s) exists
+     */
+    removeAbsCosnt: function() {
```
leshghi: See above comments about constraints

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,17 +38,526 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+            // used to be .funcX
+            repSegList[repSegLen - 1].set('refEvidencePair', satVal); 
+        }
+    }, 
+    /**
+     * Deleted Functions
+     * createID() 
+     */
 });
+IntentionBBM.numOfCreatedInstances = 0; // static variable to keep track of number of instances
```
leshghi: I think we ideally want to remove this

leshghi (commented):
This mostly looks good to me, there is some stuff that needs to be reworked that I commented but mostly restructuring things. I think instead of going through and trying to copy over more functions, a better (and ultimately more straightforward) use of your time would be to go into ElementInspector.js and try to go through the functions there one by one and re-implement them using the cell and intentionBBM. Then, as you do that you can see what Intention class functions are called/why/what they are supposed to do. From there, you should be able to rewrite for IntentionBBM as needed. That way you can avoid working on moving functions into IntentionBBM that ultimately won't be used.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
angelac2023: getIndexLastMarkedVal: function(index) { 
        var len = this.functionSegList.length; 
        if (len > 0 && index <= len) { 
            return this.functionSegList[len - index].get('refEvidencePair'); 
        }
    }

leshghi: Maybe getIndexVal?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
angelac2023: repAbsTime =  parseInt(time2) - parseInt(time1);

angelac2023: Will probably have to discuss this with @i-m-foster

leaf-ui/rappid-extensions/BackboneModelsIntention.js
angelac2023: > Can we get a comment explaining why this is the index of the first repeating segment

I think the comment was changed to "index of the last repeating segment".

angelac2023: > Can we get a comment explaining why this is the index of the first repeating segment

I think we have to discuss this further to determine if the functions actually work before adding in the comments you requested.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
angelac2023: > Add explanation of why here as well

See above reply.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
angelac2023: > Maybe getIndexVal?

I changed it to getNthLastMarkedVal with n as the parameter instead, if that works.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -13,6 +13,7 @@ var FunctionSegmentBBM = Backbone.Model.extend({
         this.type = options.type;           // Atomic function types.
```
amgrubb, line 25: Can you add `@param` for each of these variables. And all parameters for constructors and functions. Let's start adding good documentations. I like the example of `setRepeatingFunction`.

amgrubb: Same for `EvolvingFunctionBBM`.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,512 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+    
+    // TODO update all of the function descriptions
+
+    /**
+     * Returns the 4 digit representation for this
+     * EvolvingFunction's nth last function segment's
+     * marked value value. If there are no function segments, the function
+     * does not return anything.
+     */
+    getNthLastMarkedVal: function(n) {
```
amgrubb: We need to rename this: `getNthRefEvidencePair` Where is it used? If it is only used once, this may be overkill.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,512 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     *   second relative time point
+     */
+    // TODO fix this function
+    // I think we should set it so that the repeating function starts at time 1 and ends at 
+    // time 2
+    setRepeatingFunction: function(time1, time2) {
+
+        this.removeRepFuncSegments();
+        hasRepeat = true;                        
+        repAbsTime =  parseInt(time2) - parseInt(time1);
```
amgrubb: The logic on this is incorrect. Let's meet to review.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
amgrubb: The logic for this function seems off, we should probably clarify the purpose and where it is used.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,512 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     * If a RepFuncSegment does not exist in functionSegList, this function
+     * does nothing
+     */
+    removeRepFuncSegments: function() {
+        hasRepeat = false;
+        repStart = null;
+        repStop = null;          
+        repCount = null;         
+        repAbsTime = null;  
+        // Original Function
```
amgrubb: Kindly remove this commented out code.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
amgrubb: I think we want `nodeType: null`

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,512 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
         evolvingFunc = this.get('evolvingFunction');
         if (evolvingFunc != null){
             return evolvingFunc.get('functionSegList');
         }
         return null;
-    }
+    },
+ 
+    //will likely have to change this function 
+    changeInitialSatValue: function(initValue) {
```
amgrubb, line 179: Can you document the use case for this function?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,512 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        var intentionEval = analysisRequest.getUserEvaluationByID(this.get('id'), '0');
+        if (typeof intentionEval == 'undefined'){
+            return '(no value)';
+        } else {
+            return intentionEval.evaluationValue;
+        }
+    }, 
+ 
+    /**
+     * Resets dynamic function
```
amgrubb: evolving functions

leaf-ui/rappid-extensions/BackboneModelsIntention.js
amgrubb: agreed

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,512 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        this.evolvingFunction.get('id') = this.get('id'); 
+    }, 
+ 
+    /**
+     * Clears all FuncSegments for this Intention's
+     * EvolvingFunction and adds new FuncSegments according to the current
+     * function type.
+     */
+    setEvolvingFunction: function(funcType) {
+        this.evolvingFunction.set('type', funcType);
```
amgrubb, line 348: TODO: Alicia Start here in next review.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,470 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        this.evolvingFunction = new EvolvingFunctionBBM(); 
+        // Set evolving function id to intention's id 
+        this.evolvingFunction.get('id') = this.get('id'); 
+    }, 
+ 
+    /**
+     * Clears all FuncSegments for this Intention's
+     * EvolvingFunction and adds new FuncSegments according to the current
+     * function type.
+     */
```
amgrubb, line 234: Kindly, add comments how this gets called.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: It is used twice, do you think we should keep it or get rid of it?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+ 
+        var initValue = graph.get('userEvaluationList').set('id', '0').get('assignedEvidencePair');
+ 
+        // All instances of FuncSegment have been changed to FunctionSegmentBBM and the initialization process has 
+        // also been changed accordingly 
+        if (funcType == 'C' || funcType == 'R' || funcType == 'I' || funcType == 'D' || funcType == 'UD') {
+            if (funcType == 'C') {
+                var seg =  new FunctionSegmentBBM({type: funcType, refEvidencePair: initValue, startTP: '0', startAT: 0});  
+            } else if (funcType == 'R') {
+                // the marked value for a Stochastic function is always 0000
```
amgrubb: marked value -> reference evidence pair

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+            } else if (funcType == 'I' || funcType == 'D') {
+                var seg =  new FunctionSegmentBBM({type: funcType, refEvidencePair: null, startTP: '0', startAT: 'Infinity'}); 
+            } else if (funcType == 'UD') {
+                var seg =  new FunctionSegmentBBM({type: 'C', refEvidencePair: initValue, startTP: '0', startAT: 'A'}); 
+            }
+            this.getFuncSegments().push(seg);
+        } else if (funcType == 'RC' || funcType == 'CR' || funcType == 'MP' || funcType == 'MN' || funcType == 'SD' || funcType == 'DS') {
+            if (funcType == 'RC') {
+                // Stochastic and Constant
+                var seg1 =  new FunctionSegmentBBM({type: 'R', refEvidencePair: '0000', startTP: '0', startAT: 'A'});
```
amgrubb: I wonder if we could change these `'0000'` to `(no value)`?
That way null values can be edited. It might simplify the logic?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
amgrubb: Yes, I agree. This function is incorrect.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+ 
+     * Also adds a corresponding Constraint object reprenting an absolute constraint
+     * into the global model variable.
+     *
+     *This function should only be used to add new function
+     * segments for user defined functions
+     *
+     * @param {String} funcType
+     *   ex: 'C'
+     * @param {String} satValue
```
amgrubb: satValue -> refEvidencePair

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+     *   ex: '0000'
+     */
+    addUserDefinedSeg: function(funcType, satValue){
+ 
+        var len = this.evolvingFunction.getFuncSegments().length;
+        // originally used to be funcStop 
+        var start = this.evolvingFunction.getFuncSegments()[len - 1].get('stopTP');
```
amgrubb: stopTP has been removed.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        var code = start.charCodeAt(0) + 1;
+        var stop = String.fromCharCode(code);
+ 
+        //create new funcsegment model 
+        var new_model = new FunctionSegmentBBM({type: funcType, refEvidencePair: satValue, startTP: start, stopTP: stop}); 
+        this.evolvingFunction.getFuncSegments().push(new_model); 
+ 
+        model.constraints.push(new ConstraintBBM({type: 'A', srcID: this.get('id'), srcRefTP: start, destID: null, destRefTP: null}));
```
amgrubb: This is incorrect. We do not want to add a constraint at this point.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+ 
+    }, 
+    /**
+     * Sets the marked value for the FuncSegments in the
+     * EvolvingFunction for this Intention
+     *
+     * This function will only be called for I, D, RC, MP, MN functions
+     */
+    setMarkedValueToFunction: function(satValue) {
```
amgrubb, line 373: This will need to be changed/removed because we are getting rid of the associated drop down box.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        }
+    }, 
+ 
+    /**
+     * Sets the function type and marked value for the
+     * last FuncSegment for this Intention's EvolvingFunction
+     *
+     * @param {String} funcValue
+     */
+    setUserDefinedSegment: function(funcValue) {
```
amgrubb, line 395: Should this be updated and incorporated into `setEvolvingFunction`?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -39,29 +39,468 @@ var EvolvingFunctionBBM = Backbone.Model.extend({
...
+        var functionSegment = this.getFuncSegments()[funcSegLen - 1];
+        functionSegment.set('type', funcValue); 
+        if (funcValue == 'C' || funcValue =='R') {
+            funcSeg.set('refEvidencePair', '0000');
+        } 
+    }, 
+ 
+    /**
+     * Sets the satisfaction value for the last function segment
+     * in this Intention's evolving function, to satVal
```
amgrubb: Will this be part of the Function Segment BBM?

leaf-ui/js/initializeElements.js
```diff
@@ -80,6 +80,11 @@ var analysisResult = new AnalysisResult();
 
 
 var linkInspector = new LinkInspector();
+<<<<<<< HEAD
```
amgrubb: Please resolve the conflicts in this file.

amgrubb (commented):
Thanks, I finished reviewing the current state. Please update and resolve my requested changes.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
YesugenB: Should we change the` '0000' ` to `(no value)` in all of the functions (such as `setUserDefinedSegment`) or just in `setEvolvingFunction`?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
YesugenB: This function was never called so we just deleted it..

PR #439, opened 2021-06-22
Fix config highlights

### LINKED ISSUE #407: If a currently selected config is deleted, the analysisInspector and analysisConfig sidebars do not update properly by selecting a new config

YesugenB:
There is one issue (that I've noticed) where if you select a middle config but delete a different one, you will get two highlights. Will fix later today, just wanted to make an initial pr.


YesugenB:
In backbone-refactor branch as well as this branch, analysis sidebar does not update/change when config highlights do

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -124,11 +124,14 @@ var Config = Backbone.View.extend({
 
     events: {
         'click .config-elements': 'switchConfig',
-        'click .deleteconfig-button': 'removeConfig',
         'click .dropdown-button' : 'toggleDropdown',
         'dblclick .config-elements': 'rename',
         'blur .config-input': 'setConfigName',
-        'keyup .config-input': 'checkForEnter'
+        'keyup .config-input': 'checkForEnter',
+        'click .deleteconfig-button': function(e) {
```
leshghi: Okay this is awesome I didn't realize this was possible. Please share out to the rest of lab because it might be useful in other areas

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -167,7 +170,18 @@ var Config = Backbone.View.extend({
...
-        this.model.destroy();
+        index = this.model.collection.indexOf(this.model);
+        if (this.model.collection.length != 1){
+            this.model.collection.at(this.model.collection.length - 1).set({selected:false}); // Remove original highlight
+            if (index != 0){
+                this.model.collection.at(0).set({selected:true});
+            } else {
+                this.model.collection.at(1).set({selected:true});
+            }
+            this.model.destroy();
```
leshghi: Can remove the else and just put this.model.destroy() outside the if statement

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -124,11 +124,14 @@ var Config = Backbone.View.extend({
...
     events: {
         'click .config-elements': 'switchConfig',
-        'click .deleteconfig-button': 'removeConfig',
         'click .dropdown-button' : 'toggleDropdown',
         'dblclick .config-elements': 'rename',
         'blur .config-input': 'setConfigName',
-        'keyup .config-input': 'checkForEnter'
+        'keyup .config-input': 'checkForEnter',
+        'click .deleteconfig-button': function(e) {
+            this.switchConfig(e);
```
leshghi: We don't want to call switchConfig here though, because that will switch the deleted config to being selected

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -167,7 +170,18 @@ var Config = Backbone.View.extend({
      * and triggering a removal of its respective view 
      */
     removeConfig:function(){
-        this.model.destroy();
+        index = this.model.collection.indexOf(this.model);
```
leshghi, line 172: Thoughts on this logic?
if(model is selected && length of configCollection > 1)
     if(model index > 0) -> set model at index-1 to selected = true
     else -> set model at index + 1 to select = true
destroy model

leaf-ui/rappid-extensions/ConfigInspector.js
YesugenB: I called it there because without the switchConfig in line 132, there were multiple configs highlighted as the previously highlighted config was not being unhighlighted. Do you think I should try to use another switchConfig or something similar elsewhere or remove the original switchConfig and figure something else out?

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -167,7 +170,18 @@ var Config = Backbone.View.extend({
...
+        if (this.model.set({selected:true}) && this.model.collection.length > 1){
+           if (index > 0 ){
+                this.model.collection.at(index - 1).set({selected:true});
+                this.model.collection.at(index - 1).trigger('change:switchConfigs', this.model.collection.at(index-1)); 
+           } else {
+                this.model.collection.at(index + 1).set({selected:true});       
+                this.model.collection.at(index + 1).trigger('change:switchConfigs', this.model.collection.at(index+1)); 
+            }
+           this.model.destroy();
```
leshghi: This isn't needed since you call it outside of the if statement

leaf-ui/rappid-extensions/ConfigInspector.js
leshghi: You can remove switchConfig(e) from here now since you check to see if the model is selected before switching to another model in the remove function

leaf-ui/rappid-extensions/ConfigInspector.js
leshghi: We don't need the e since the event is never used

leshghi (approved):
Looks good just some minor changes. Also - you'll need to add some code so that when you delete a selected config and the selection switches over the analysis parameters switch to the newly selected config as well. **Make an issue** for that though instead of working on it here since I know @angelac2023 and @Manarnz are working on something similar and I don't want you guys to have merge conflicts - So this is fine for now and then once they are done you or they can tackle the switching issue with the new structure in place.

PR #441, opened 2021-06-23
Models restructure as part of Configurations revamp - AnalysisModels.js

angelac2023:
Please let us know if there are any changes to be made.


leaf-ui/rappid-extensions/AnalysisModels.js
```diff
@@ -13,9 +13,7 @@ var ResultModel = Backbone.Model.extend({
     idAttribute: "uid",
 
     defaults: {
-        name:"Default Result",
```
leshghi, line 16: Probably still needs the name and selected params for now to allow you to switch between results and have a name that renders on the sidebar

leaf-ui/rappid-extensions/AnalysisModels.js
```diff
@@ -92,6 +85,25 @@ var ConfigModel = Backbone.Model.extend({
     },
 });
 
+/**
+ * Analysis Parameters 
+ */
+
+var ConfigAnalysis = Backbone.Model.extend ({ 
+    initialize : function(options){
```
leshghi: I think we might need the action param in here too @amgrubb can you confirm?

leaf-ui/rappid-extensions/AnalysisModels.js
```diff
@@ -92,6 +85,25 @@ var ConfigModel = Backbone.Model.extend({
     },
 });
 
+/**
+ * Analysis Parameters 
+ */
+
+var ConfigAnalysis = Backbone.Model.extend ({
```
leshghi:
```suggestion
var AnalysisParametersBBM = Backbone.Model.extend ({ 
```

leaf-ui/rappid-extensions/AnalysisModels.js
```diff
@@ -92,6 +85,25 @@ var ConfigModel = Backbone.Model.extend({
...
+/**
+ * Analysis Parameters 
+ */
+
+var ConfigAnalysis = Backbone.Model.extend ({ 
+    initialize : function(options){
+        _.extend({}, this.defaults, options); 
+    },
+
+    idAttribute: "uid",
```
leshghi:
```suggestion
```

leaf-ui/rappid-extensions/AnalysisModels.js
leshghi: I also think its possible we may want the Result to hold the AnalysisParametersBBM. @amgrubb thoughts?

leshghi (changes requested):
Some naming convention changes and moving things around. I only looked at the model restructure file because that is what we are reviewing. The buttons seem to be in progress and I think can wait/be separated from this PR.

leaf-analysis/app.js
```diff
@@ -4,7 +4,7 @@
 
 // Name of .jar file for BloomingLeaf project must be Blooming.jar
 //var userPath = "/Users/<your user path here>/BloomingLeaf"
-var userPath = "/Users/judySmith/git/BloomingLeaf"
+var userPath = "/Users/ChuAn_nu0wj7q/Desktop/BloomingLeaf"
```
amgrubb, line 7: Please remove this change.

leaf-ui/index.html
```diff
@@ -170,13 +184,16 @@
         <script src="./rappid-extensions/BackboneModelsOther.js"></script>
         <script src="./rappid-extensions/AnalysisModels.js"></script>
         <script src="./rappid-extensions/AssignmentsTable.js"></script>
+        <!-- new addition -->
```
amgrubb: You can remove this line.

leaf-ui/rappid-extensions/AnalysisModels.js
amgrubb: Yes, we need a default name, for now.
Also, we discussed having default analysis parameters and then when the analysis is called, saving the results with the analysis parameters.

amgrubb (commented):
I'll take another look another day.

leaf-ui/rappid-extensions/AnalysisModels.js
amgrubb: @leshghi you mean to distinguish between single path and next states? Yes, we should add the action parameter.

amgrubb: Good catch.

amgrubb:
Replaced by #461

PR #442, opened 2021-06-23
Implemented the Embedding of Intentions in Actor

Domily11:
Took out embedBasicActior and determined that intentionID is not needed as a list of embedded cells are automatically generated

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -751,7 +751,8 @@ var element_counter = 0;
 
 // Whenever an element is added to the graph
 graph.on("add", function(cell) {
-	if (cell instanceof joint.dia.CellLink){
+    var createdInstance = paper.findViewsInArea(paper.getArea())  
+	if (cell instanceof joint.dia.Link){
         if (graph.getCell(cell.get("source").id) instanceof joint.shapes.basic.Actor){
             cell.prop("type", "Actor");
             cell.label(0,{attrs:{text:{text:"is-a"}}});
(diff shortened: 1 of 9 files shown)


leaf-ui/js/onFunctions.js
```diff
@@ -891,22 +891,32 @@ paper.on({
...
                             var userIntention = model.getIntentionByID(cell.attributes.nodeID);
                             if (userIntention.nodeActorID !== '-') {
                                 var actor = model.getActorByID(userIntention.nodeActorID);
                                 actor.removeIntentionID(userIntention.nodeID);
                             }
+                            */
                         }
                         // embed element in new actor
-                        embedBasicActor(cell);
+                        var overlapCells = paper.findViewsFromPoint(cell.getBBox().center());
```
leshghi, line 879: Can you reset the IntentionBBM actorID variable to '-' if overlapCells.length = 0

leshghi (changes requested):
Looks good to me - respond to comment and pull from backbone-refactor again and you should be set!

leaf-ui/js/onFunctions.js
Domily11: You don't need it tho. cell.get('parent') automatically gets the node the cell is embedded to

Domily11: If it is unembedded the function doesn't return the parent

leshghi: Oh awesome - should we remove the ActorId param from intentions then?

Domily11: I think so

leshghi (approved):
Looks good for just merging into another one of your branches. Will do a more comprehensive review for mack-actorinspector PR

Issue #445, opened 2021-06-23
LinkInspector EvolvingLink select for elements is only dynamic one way


leshghi:
When you start from the begin(linkType) and then chose the end(postType). However, if you then change the end(postType) the begin dropdown does not update with the correct options

PR #455, opened 2021-06-24
Fixed Issue #445

Domily11:


leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -272,6 +278,7 @@ var LinkInspector = Backbone.View.extend({
             this.link.set('linkType', linkType);
             this.link.set('postType', postType);
             this.model.label(0, {position: 0.5, attrs: {text: {text: linkType + " | " + postType}}});
+            $('#link-type-end').val(this.link.get('postType'));
```
leshghi: I know its easier/more concise to put this line in here but since this function is really designed to alter cell properties, not html properties, I think it would be better to put it after you call setValues in each of the if statements.

leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -201,8 +201,14 @@ var LinkInspector = Backbone.View.extend({
         $('option').show(); // Clear the previous selection
         if (begin == 'and' || begin == 'or'){
             $('option.B').hide(); // Hide options incompatible with and/or selection such as +,-,++S, etc
+            if($('#link-type-end :selected').attr('class') == 'B'){
```
leshghi: Add comments for these if statements

leshghi (commented):
This looks great - two things just about moving stuff around and commenting. Also add in a line or two description to your pull request about what the issue is. Finally for a slight feature add - this is slightly outside the scope of the issue but could you make it so that if the end select is 'and', if you select 'and' from the begin select it resets the end to be a placeholder/null? Aka if you select from begin the same option that end has, end resets.

leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -213,6 +222,10 @@ var LinkInspector = Backbone.View.extend({
     updateEndEvolRelations: function() {
         // Save based on evolving relations
         this.setValues(this.link.get('linkType'), $('#link-type-end').val(), true)
+
+        if(this.link.get('postType')!=null){
+            $('#link-type-begin option[value= \'' + this.link.get('postType') + '\']').hide(); // Hide the selected postType value in begin so that user can't duplicate
```
leshghi:
```suggestion
            $('#link-type-begin option[value= \'' + this.link.get('postType') + '\']').hide(); // Hide the selected postType value in begin dropdown so that user can't select the same value for the linkType and postType
```

leaf-ui/rappid-extensions/LinkInspector.js
```diff
@@ -201,10 +201,19 @@ var LinkInspector = Backbone.View.extend({
         $('option').show(); // Clear the previous selection
         if (begin == 'and' || begin == 'or'){
             $('option.B').hide(); // Hide options incompatible with and/or selection such as +,-,++S, etc
+            if($('#link-type-end :selected').attr('class') == 'B'){ //if user later chose an option in the beginning but the end is incompatible with and/or (ex. +,-,++S)
```
leshghi: For all these comments can you try to make what they are doing a bit clearer, (aka - if end selection is incompatible with new begin selection, clear end). And use the format // Space and capital letter to start the comment

leshghi (approved):
Looks good, only comments are minor comment stuff so preapproving and once you make those changes you are set to squash and merge. Awesome job thanks!

PR #459, opened 2021-06-28
Implement IntentionBBM into ElementInspector.js

i-m-foster:


leaf-ui/js/object/modelObjects.js
```diff
@@ -1116,6 +1116,7 @@ class IntentionTest {
      */
     constructor(nodeActorID, nodeType) {
         this.nodeActorID = nodeActorID;
+        //this for some reason is not updating??
```
leshghi: Remove this file

leaf-ui/js/onFunctions.js
```diff
@@ -727,21 +727,14 @@ $('#colorblind-mode-isOn').on('click', function(){ //turns off colorblind mode
...
+    cell.set('intention', newIntentionBBM);
+    cell.attr('.funcvalue/text', ' ');
 
     // create intention evaluation object
-    var intentionEval = new UserEvaluation(intention.nodeID, '0', '(no value)');
-    analysisRequest.userAssignmentsList.push(intentionEval);
-
-    cell.attributes.nodeID = intention.nodeID;
 
+    var intentionEval = new UserEvaluationBBM({intentionID: newIntentionBBM.cid, absTP: '0', assignedEvidencePair: '(no value)'});
```
leshghi: You can combine this into one line since you don't use intentionEval anywhere else. Also - could the assigned evidence pair be in defaults on the UserEvaluationBBM?

leaf-ui/js/onFunctions.js
```diff
@@ -764,8 +757,9 @@ graph.on("add", function(cell) {
             cell.set('link', new LinkBBM({}));
         }
     } else if (cell instanceof joint.shapes.basic.Intention){
-		cell.set('intention', new IntentionBBM({}));
-		cell.attr('.funcvalue/text', ' ');
+        createIntention(cell);
+		// cell.set('intention', new IntentionBBM({}));
```
leshghi: Remove comments

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+            repStart: null,         // 0, A, B, ..
+            repStop: null,          // B, C, D, ..
+            repCount: null,         // 2, 3, ..n
+            repAbsTime: null,       // 0, 1, ..n
+        }
        
     },
     
-    //TODO: I think could have some helper functions for the constructor. 
+    //TODO: I think could have some helper functions for the constructor.
```
leshghi: Remove comment

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+     * @param {Integer} count
+     *  Number of times repeated segment is repeated 
+     * @param {Integer} absTime
+     *  Absolute time of repeating segment 
+     */
+    setRepeatingFunction: function(start, stopRep, count, absTime) {
+
+        this.removeRepFuncSegments();
+        hasRepeat = true;                         
+        repStart = start;
```
leshghi: I could be wrong but shouldn't these parameters be accessed by "this.hasRepeat"? Also - if you are setting all the parameters in this function, you don't need to call this.removeRepFuncSegments()

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
     /**
-     * @returns Array of FunctionSegmentBBMs, or null if it is not an evolvingFunction
-     */
+    * @returns Array of FunctionSegmentBBMs, or null if it is not an evolvingFunction
+    */
     getFuncSegments: function(){
+
         evolvingFunc = this.get('evolvingFunction');
         if (evolvingFunc != null){
+            console.log(evolvingFunc.get('functionSegList'));
```
leshghi: remove console log

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
-    }
+    },
+ 
+    /**
+     * Called whenever the user changes or removes the Initial Satisfaction Value
+     * in the Element Inspector. Called on Change for #init-sat-value 
+     * 
+     * Sets the refEvidence pair based on the chosen initValue
+     * 
+     * @param {*} initValue
```
leshghi: Put in param type in the brackets here

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+        intentionEval.set('assignedEvidencePair', initValue);
+        var funcSegList = this.getFuncSegments();
+        
+        if (this.get('evolvingFunction') != null) {
+            if (this.get('evolvingFunction').get('type') == 'C' || 
+                (this.get('evolvingFunction').get('type') == 'UD' && funcSegList[0].get('type') == 'C')) { 
+                    funcSegList[0].set('refEvidencePair', initValue); // Set first index of funcSegList to given initValue 
+            }
+            this.get('evolvingFunction').set('type', 'NT');
+            this.get('evolvingFunction').set('functionSegList', []);
```
leshghi, line 225: I'm confused - why are we setting the first value in the funcSegList just to then set the entire list to an empty array?

Also - this function needs more comments

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+        this.changeInitialSatValue('(no value)');
+    },    
+ 
+    /**
+     * Returns the 4 digit representation for the
+     * Intention's initial satisfaction value
+     *
+     * @returns {String}
+     */
+    getInitialSatValue: function() {
```
leshghi: Is this different than the initialValue stored on the intention?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+            return intentionEval.get('assignedEvidencePair'); 
+        }
+    }, 
+ 
+    /**
+     * Resets evolving functions
+     */
+    removeFunction: function() {
+
+        this.removeAbsConstraint();
```
leshghi: So - we are now storing absolute constraints directly on the FuncSegmentBBM. The startAT represents the old absTime on the constraint, and is the only thing that should need to be updated/changed when you add/remove/change an absolute constraint. I'd check in with @amgrubb to chat more about this

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+     * function type.
+     * Called when user changes the function type in Element Inspector 
+     * @param {String} funcType
+     */
+    setEvolvingFunction: function(funcType) {
+
+        this.set('evolvingFunction', new EvolvingFunctionBBM({type: funcType, functionSegList: []}));
+
+        // Since function changed, remove all current absolute constraints related to this intention
+        this.removeAbsConstraint();
```
leshghi: Can remove this since clearing the functionSegList removes constraints stored on FuncSegBBMs

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+ 
+        // Add new absolute constraints if required
+        this.addAbsConstraint(funcType);
```
leshghi: I don't think I understand the use case for this function - why add an absolute constraint if you don't have a time point? @amgrubb ?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+ 
+        // var initValue = graph.get('userEvaluationList').get(this.cid, '0').get('assignedEvidencePair');
```
leshghi: Remove comment

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+        var initValue = graph.getUserEvaluationBBM(this.cid, '0').get('assignedEvidencePair');
```
leshghi: Again is this different than the intention initial value? It totally may be I just am not sure

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+
+        // Creates the correct FunctionSegmentBBM(s) for the selected function type
+        if (funcType == 'C' || funcType == 'R' || funcType == 'I' || funcType == 'D' || funcType == 'UD') {
+            if (funcType == 'C') {
+                var seg =  new FunctionSegmentBBM({type: funcType, refEvidencePair: initValue, startTP: '0', startAT: 0});  
+            } else if (funcType == 'R') {
+                // The reference evidence pair for a Stochastic function is always 0000
+                var seg =  new FunctionSegmentBBM({type: funcType, refEvidencePair: '(no value)', startTP: '0', startAT: 0}); 
+            } else if (funcType == 'I' || funcType == 'D') {
+                var seg =  new FunctionSegmentBBM({type: funcType, refEvidencePair: null, startTP: '0', startAT: 0});
```
leshghi: Why is the refEvidencePair null for this one?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+                // Stochastic and Constant
+                var seg1 =  new FunctionSegmentBBM({type: 'R', refEvidencePair: '(no value)', startTP: '0', startAT: 0}); 
+                var seg2 =  new FunctionSegmentBBM({type: 'C', refEvidencePair: null, startTP: 'A', startAT: myNull}); 
+            } else if (funcType == 'CR') {
+                // Constant and Stochastic
+                var seg1 =  new FunctionSegmentBBM({type: 'C', refEvidencePair: initValue, startTP: '0', startAT: 0}); 
+                var seg2 =  new FunctionSegmentBBM({type: 'R', refEvidencePair: '(no value)', startTP: 'A', startAT: myNull}); 
+            } else if (funcType == 'MP') {
+                // Increase and Constant
+                var seg1 =  new FunctionSegmentBBM({type: 'I', refEvidencePair: null, startTP: '0', startAT: 0});
```
leshghi, line 284: Same question about why the first segment has a null refEvidencePair - isn't always the initial value for TP 0? Could be misunderstanding tho

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+     * representing an absolute constraint, if requried.
+     *
+     * If funcType is RC, CR, MP, MN, SD, DS  a Constraint object
+     * representing an absolute constraint will be added. If not, this
+     * function does not do anything
+     *
+     * @param {String} funcType
+     *   ex: 'RC'
+     */
+    addAbsConstraint: function(funcType) {
```
leshghi: I think you should be able to delete this

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+            graph.get('constraints').push(new ConstraintBBM({type: 'A', srcID: this.cid, srcRefTP: 'A', destID: null, destRefTP: null})); 
+        }
+    }, 
+ 
+    /**
+     * Removes the absolute Constraint object(s) for the Intention from
+     * the global graph variable, if such absolute Constraint object(s) exists
+     */
+
+    removeAbsConstraint: function() {
```
leshghi: Think you should be able to delete this too, at least as written

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+     * @param {String} funcType
+     *   ex: 'C'
+     * @param {String} refEvidencePair
+     *   ex: '0000'
+     * @param {Integer} startTime
+     */
+    addUserDefinedSeg: function(funcType, refEvidencePair, startTime){
+
+        var len = this.getFuncSegments().length;
+        var startCheck = this.getFuncSegments()[len - 1].get('startTP'); // Get last value in list
```
leshghi, line 323: Are you allowed to use -1 to access last value in the list? Unsure but would check because it would be more efficient

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+     *   ex: '0000'
+     * @param {Integer} startTime
+     */
+    addUserDefinedSeg: function(funcType, refEvidencePair, startTime){
+
+        var len = this.getFuncSegments().length;
+        var startCheck = this.getFuncSegments()[len - 1].get('startTP'); // Get last value in list 
+        if (startCheck == '0') { // If previous segment is at 0 then next one is at A
+            var start = 'A';
+        } else { // Otherwise, for following segment increase number by one
```
leshghi: Increase number or letter?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+        if (startCheck == '0') { // If previous segment is at 0 then next one is at A
+            var start = 'A';
+        } else { // Otherwise, for following segment increase number by one 
+            var start = String.fromCharCode(startCheck.charCodeAt(0) + 1); 
+        }
+ 
+        // Create new FunctionSegmentBBM and adds it to functionSegList 
+        var new_model = new FunctionSegmentBBM({type: funcType, refEvidencePair: refEvidencePair, startTP: start, startAT: myNull}); 
+        this.getFuncSegments().push(new_model); 
+        //graph.get('constraints').push(new ConstraintBBM({type: 'A', srcID: this.get('cid'), srcRefTP: start, destID: null, destRefTP: null}));
```
leshghi: Can remove this line

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+        var startCheck = this.getFuncSegments()[len - 1].get('startTP'); // Get last value in list 
+        if (startCheck == '0') { // If previous segment is at 0 then next one is at A
+            var start = 'A';
+        } else { // Otherwise, for following segment increase number by one 
+            var start = String.fromCharCode(startCheck.charCodeAt(0) + 1); 
+        }
+ 
+        // Create new FunctionSegmentBBM and adds it to functionSegList 
+        var new_model = new FunctionSegmentBBM({type: funcType, refEvidencePair: refEvidencePair, startTP: start, startAT: myNull}); 
+        this.getFuncSegments().push(new_model);
```
leshghi: You can inline the new FunctionSegmentBBM line into the this.getFuncSegments().push() line because you don't use the variable anywhere else

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+    // TODO: i think we are going to have to change this function when we add views for the FunctionSegmentBBMs
+    setMarkedValueToFunction: function(satValue) {
+
+        if (this.get('evolvingFunction') != null) {
+            var funcType = this.get('evolvingFunction').get('type'); 
+        } else { 
+            var funcType = null; 
+        }
+ 
+        var len = this.getFuncSegments().length;
```
leshghi, line 349: What if evolving function is null here? Then this line would throw an error

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+        var new_model = new FunctionSegmentBBM({type: funcType, refEvidencePair: refEvidencePair, startTP: start, startAT: myNull}); 
+        this.getFuncSegments().push(new_model); 
+        //graph.get('constraints').push(new ConstraintBBM({type: 'A', srcID: this.get('cid'), srcRefTP: start, destID: null, destRefTP: null}));
+ 
+    }, 
+    /**
+     * Sets the refEvidencePair for the FunctionSegmentBBMs in the
+     * EvolvingFunction for the Intention
+     *
+     * This function will only be called for I, D, RC, MP, MN functions
```
leshghi: So I notice for MP this will set the 'refEvidencePair' for both func segments to the satValue - is that the intended behavior?

leshghi: Also - what happens if you want to set function type and marked value for something other than the last func segment for MP and MN? Like if you change the beginning value after already setting an end one

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+     * @param {String} funcValue
+     */
+    setUserDefinedSegment: function(funcValue) {
+
+        if (this.getFuncSegments() != null) {
+            var funcSegLen = this.getFuncSegments().length;
+        } else { 
+            var funcSegLen = null;
+        }
+        var functionSegment = this.getFuncSegments()[funcSegLen - 1];
```
leshghi: so here you might be calling null-1 which would throw an error

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,404 @@
...
+ 
+    /**
+     * Sets the function type and marked value for the
+     * last FunctionSegmentBBM for the Intention's EvolvingFunction
+     *
+     * @param {String} funcValue
+     */
+    setUserDefinedSegment: function(funcValue) {
+
+        if (this.getFuncSegments() != null) {
```
leshghi, line 364: logic:
var funcSegments = this.getFuncSegments()
if (funcSegments != null){
get length and do all your stuff in here
} no else needed

leaf-ui/rappid-extensions/BackboneModelsOther.js
```diff
@@ -66,6 +66,7 @@ var UserEvaluationBBM = Backbone.Model.extend({
 
 var UserEvaluationCollection = Backbone.Collection.extend({
     model: UserEvaluationBBM
+
```
leshghi: Remove this file

leaf-ui/rappid-extensions/IntermediateValues.js
```diff
@@ -483,11 +483,13 @@ var IntermediateValuesTable = Backbone.View.extend({
     // Needed after looping through and appending rows to avoid voiding the select
     $('.intention-row').each(function () {
         $(this).find('select').each(function () {
-            var nodeID = $(this).attr('nodeID');
+            //var nodeID = $(this).attr('nodeID');
```
leshghi: Remove this file from commit

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -139,20 +143,21 @@ var ElementInspector = Backbone.View.extend({
...
             '<canvas id="chart" width="240" height="240"></canvas>',
-
-    ].join(''),
+            
+            '</script>'].join(''),
 
     events: {
         'change #init-sat-value':'initSatValueChanged',
-
+        'click .intention-elements': 'switchModel',
```
leshghi: There is no intention elements class, and I don't think you should need to switch the model from inside the view

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -165,21 +170,17 @@ var ElementInspector = Backbone.View.extend({
...
         'click #constraint-restart': 'removeUserConstraints',
-        'keyup .cell-attrs-text': 'nameAction'
+        'keyup .cell-attrs-text': 'nameAction',
+        'clearInspector .inspector-views' : 'removeView',
     },
 
     /**
      * Initializes the element inspector using previously defined templates
      */
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -191,22 +192,27 @@ var ElementInspector = Backbone.View.extend({
         this.userConstraintsHTML = $("#new-user-constraints").last().clone();
 
         // Load initial value and node name
-        this.$('.cell-attrs-text').val(this.intention.nodeName);
-        this.$('#init-sat-value').val(satisfactionValuesDict[this.intention.getInitialSatValue()].name);
+        this.$('.cell-attrs-text').val(this.model.get('intention').get('nodeName'));
```
leshghi: If you use this.model.get('intention') a lot - I would recommend in initialize putting a line "this.intention = this.model.get('intention')" so that you can use this.intention to access the intention throughout the code

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -191,22 +192,27 @@ var ElementInspector = Backbone.View.extend({
...
         this.repeatOptionsDisplay = false;
         // TODO what does this do?
         this.setRepeatConstraintMode("TurnOff");
 
         // Load initial value for function type in the html select element
-        var functionType = this.intention.dynamicFunction.stringDynVis;
+        if (this.model.get('intention').get('evolvingFunction') != null ) {
+            var functionType = this.model.get('intention').get('evolvingFunction').get('type');
+        }
+        else {var functionType = null;}
```
leshghi: If you don't want to do anything when evolving function is null, you can just put all of the below code into the if (this.model.get('intention').get('evolvingFunction') != null ) and remove the else

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -219,38 +225,48 @@ var ElementInspector = Backbone.View.extend({
         }
 
         this.updateCell();
-        */
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -219,38 +225,48 @@ var ElementInspector = Backbone.View.extend({
...
-            this.$('.function-type').append('<option value=CR> Constant-Stochastic </option>');
-            this.$('.function-type').append('<option value=MP> Montonic Positive </option>');
-            this.$('.function-type').append('<option value=MN> Montonic Negative </option>');
-            this.$('.function-type').append('<option value=SD> Satisfied Denied </option>');
-            this.$('.function-type').append('<option value=DS> Denied Satisfied </option>');
-            this.$('.function-type').append('<option value=UD> User Defined </option>');
+    */
+  checkInitialSatValue: function() {
+
+          if (this.model.get('intention').getInitialSatValue() == '(no value)') {
```
leshghi: Thoughts on using a class that can hide or show options based on if it is Stochastic or not? Check out what was done in the LinkInspector.js and ask @Domily11 if you have questions because she did something similar there and it could clean up this function a lot

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -219,38 +225,48 @@ var ElementInspector = Backbone.View.extend({
...
+             this.$('.function-type').append('<option value=D> Decrease </option>');
+             this.$('.function-type').append('<option value=RC> Stochastic-Constant </option>');
+             this.$('.function-type').append('<option value=CR> Constant-Stochastic </option>');
+             this.$('.function-type').append('<option value=MP> Montonic Positive </option>');
+             this.$('.function-type').append('<option value=MN> Montonic Negative </option>');
+             this.$('.function-type').append('<option value=SD> Satisfied Denied </option>');
+             this.$('.function-type').append('<option value=DS> Denied Satisfied </option>');
+             this.$('.function-type').append('<option value=UD> User Defined </option>');   
         }
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -219,38 +225,48 @@ var ElementInspector = Backbone.View.extend({
...
     },
 
     nameAction: function(event) {
+
```
leshghi, line 255: remove line. also this predates you but ENTER_KEY is only used once so maybe just remove the variable and replace with 13 directly in the below if statement and then just comment what it means

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -270,6 +287,7 @@ var ElementInspector = Backbone.View.extend({
      * @returns {Object}
      */
     initializeSatValueOptions: function() {
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -316,20 +334,20 @@ var ElementInspector = Backbone.View.extend({
     /**
      * Initializes components to display user defined functions
      */
-    renderUserDefined: function(cell){
+    renderUserDefined: function(){
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -316,20 +334,20 @@ var ElementInspector = Backbone.View.extend({
...
+    renderUserDefined: function(){
+        
         this.$('#markedValue').hide();
         $(".function-type").val('UD');
 
         // Load the user defined constraints
-        var len = this.intention.getNumOfFuncSegements();
-        var funcSegments = this.intention.dynamicFunction.getFuncSegmentIterable();
-
+        var len = this.model.get('intention').getFuncSegments().length;
```
leshghi: Would it be possible to have a null evolving function at this point? Also - if you switch these lines its a lot less words.
Aka:
var funcSegments = this.model.get('intention').getFuncSegments();
var len = funcSegments.length

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -316,20 +334,20 @@ var ElementInspector = Backbone.View.extend({
...
+        var funcSegments = this.model.get('intention').getFuncSegments();
 
         for (var i = 0; i < len; i++) {
 
             // set the intial values
-            $(".user-sat-value").last().val(satisfactionValuesDict[funcSegments[i].funcX].name);
-            $(".user-function-type").last().val(funcSegments[i].funcType);
+            $(".user-sat-value").last().val(satisfactionValuesDict[funcSegments[i].get('refEvidencePair')].name);
```
leshghi: You could use an enhanced for loop here aka (syntax might be slightly off but this is the idea)
```
for (let funcSegment in funcSegments){
$(".user-sat-value").last().val(satisfactionValuesDict[funcSegment.get('refEvidencePair')].name);
etc etc etc
}
```
That way you don't have to then use i to access the funcSegment

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -343,11 +361,11 @@ var ElementInspector = Backbone.View.extend({
...
 
-        if (this.intention.dynamicFunction.hasRepeat()) {
-            var repBegin = this.intention.dynamicFunction.getStartRepeatEpoch();
-            var repEnd = this.intention.dynamicFunction.getEndRepeatEpoch();
-            var repNum = this.intention.dynamicFunction.getRepeatRepNum();
-            var absTime = this.intention.dynamicFunction.getRepeatAbsTime();
+        if (this.model.get('intention').get('evolvingFunction').get('hasRepeat')) {
+            var repBegin = this.model.get('intention').get('evolvingFunction').get('repStart');
+            var repEnd = this.model.get('intention').get('evolvingFunction').get('repStop');
+            var repNum = this.model.get('intention').get('evolvingFunction').get('repCount');
```
leshghi: Could you just inline these in the below lines instead of creating variables for them since they are only used once?

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -361,6 +379,7 @@ var ElementInspector = Backbone.View.extend({
         }
 
         this.updateChartUserDefined(null);
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -370,25 +389,31 @@ var ElementInspector = Backbone.View.extend({
      * This function is called on change for #init-sat-value,
      */
     initSatValueChanged: function(event) {
```
leshghi: You don't need the event if you don't use it. Its one of the "unspoken" parameters. You don't need to explicitly have it in the params unless you want to access it. (Kind of like "options" in the initialize function if that is a helpful way to think about it")

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -370,25 +389,31 @@ var ElementInspector = Backbone.View.extend({
...
     /**
-     * Clears all FuncSegments for this intention's Intention object's
+     * Clears all FuncSegments for this intention's
      * EvolvingFunction and adds new FuncSegments according to the current
      * function type.
      *
      * This function is called on change for .function-type.
      */
     funcTypeChanged: function(event) {
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -370,25 +389,31 @@ var ElementInspector = Backbone.View.extend({
...
      * function type.
      *
      * This function is called on change for .function-type.
      */
     funcTypeChanged: function(event) {
+        
         var funcType = this.$('.function-type').val();
-        this.intention.setEvolvingFunction(funcType);
+
+        this.model.get('intention').setEvolvingFunction(funcType);
```
leshghi: Can inline this.$('.function-type').val() here

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -370,25 +389,31 @@ var ElementInspector = Backbone.View.extend({
...
+        console.log(this.$('.function-type').val());
```
leshghi: Remove console log

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -370,25 +389,31 @@ var ElementInspector = Backbone.View.extend({
...
         this.updateCell(null);
         this.updateHTML(event);
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
      * This function is called on change for .user-function-type
      */
     userFuncTypeChanged: function(event) {
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
      * This function is called on change for .user-function-type
      */
     userFuncTypeChanged: function(event) {
+        
         var funcType = this.$('.user-function-type').last().val();
-        this.intention.setUserDefinedSegment(funcType);
+        this.model.get('intention').setUserDefinedSegment(funcType);
         this.updateHTML(event);
+
```
leshghi: Remove line

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
...
+        this.model.get('intention').setUserDefinedSegment(funcType);
         this.updateHTML(event);
+        
     },
 
     /**
      * This function is called on change for .user-sat-value
      */
     userSatValChanged: function(event) {
+
```
leshghi: Remove extra lines at beginning/end of this function

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
...
     /**
      * This function is called on change for .user-sat-value
      */
     userSatValChanged: function(event) {
+        
         var satVal = satValueDict[this.$('.user-sat-value').last().val()];
-        this.intention.updateLastFuncSegSatVal(satVal);
+
+        // Sets the satisfaction value for the last function segment in
+        // the Intention's evolving function to satVal
```
leshghi:
```suggestion
        // In the Intention's evolving function to satVal
```

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
...
 
     /**
      * This function is called on change for .user-sat-value
      */
     userSatValChanged: function(event) {
+        
         var satVal = satValueDict[this.$('.user-sat-value').last().val()];
-        this.intention.updateLastFuncSegSatVal(satVal);
+
+        // Sets the satisfaction value for the last function segment in
```
leshghi:
```suggestion
        // Sets the satisfaction value for the last function segment
```

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
...
+        // the Intention's evolving function to satVal
+        var funcSegLen = this.model.get('intention').getFuncSegments().length;
+        this.model.get('intention').getFuncSegments()[funcSegLen - 1].set('refEvidencePair', satVal);
```
leshghi: Check for sure if you can use -1 here cause it would clean up the code a lot

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -397,28 +422,39 @@ var ElementInspector = Backbone.View.extend({
...
 
     /**
-     * Sets the marked value
+     * Sets the refEvidencePair for the FunctionSegmentBBMs
      */
     funcSatValChanged: function(event) {
+        
         var satValue = satValueDict[this.$('#markedValue').val()]; // 4 digit representation
-        this.intention.setMarkedValueToFunction(satValue);
+        this.model.get('intention').setMarkedValueToFunction(satValue);
```
leshghi: Can inline here too, and remove extra lines at beginning and end

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -465,7 +503,8 @@ var ElementInspector = Backbone.View.extend({
             }
         }
 
-        this.updateChart(null);
+        this.updateChart(null);
```
leshghi: You never need to pass anything into updateChart, and in fact can remove the event parameter from that function as well because it is never used for anything

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -474,6 +513,7 @@ var ElementInspector = Backbone.View.extend({
      * value or the function type accordingly.
      */
     validityCheck: function(event) {
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -509,9 +550,10 @@ var ElementInspector = Backbone.View.extend({
      * select element (#markedValue) for the associated satisfaction value.
      */
     displayFunctionSatValue: function(event) {
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -546,7 +590,8 @@ var ElementInspector = Backbone.View.extend({
         }
         
         if (func == 'I' || func == 'D') {
-            var prevVal = satisfactionValuesDict[this.intention.dynamicFunction.getSecondLastMarkedVal()].name;
+            console.log(this.model.get('intention').get('evolvingFunction'));
```
leshghi: Remove console log

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -537,6 +580,7 @@ var ElementInspector = Backbone.View.extend({
      * indicate satisfaction values when creating a user defined function.
      */
     addUDFunctionValues: function(event) {
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -562,36 +607,43 @@ var ElementInspector = Backbone.View.extend({
...
     },
 
     /**
      * Modifies the passed in datasets with their default values
      * @param {Array.<Object>}
      */
     resetChartDatasets: function(datasets) {
-
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -562,36 +607,43 @@ var ElementInspector = Backbone.View.extend({
...
     },
 
     /**
      * Updates the chart to represent data related to the the current function and
      * satisfaction value(s)
      */
     updateChart: function(event) {
-        var funcType = this.intention.dynamicFunction.stringDynVis;
-        var initVal = satisfactionValuesDict[this.intention.getInitialSatValue()].chartVal;
+
```
leshghi: Remove extra lines and console logs

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -641,43 +692,46 @@ var ElementInspector = Backbone.View.extend({
         }
 
         this.chart.display(context);
+        
     },
 
     getUDChartLabel: function(num) {
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -641,43 +692,46 @@ var ElementInspector = Backbone.View.extend({
...
         return res;
+    
     },
 
     /**
      * Updates the chart to represent data related to the the current user
      * defined function and satisfaction value(s)
      */
     updateChartUserDefined: function(event) {
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -641,43 +692,46 @@ var ElementInspector = Backbone.View.extend({
...
 
     /**
      * Updates the chart to represent data related to the the current user
      * defined function and satisfaction value(s)
      */
     updateChartUserDefined: function(event) {
+        
         var context = $("#chart").get(0).getContext("2d");
-        var numFuncSegments = this.intention.getNumOfFuncSegements();
+        var numFuncSegments = this.model.get('intention').getFuncSegments().length;
```
leshghi: This would never be null, right?

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -641,43 +692,46 @@ var ElementInspector = Backbone.View.extend({
...
-
-
         // Get init sat value
-        var initSatVal = satisfactionValuesDict[this.intention.getInitialSatValue()].chartVal;
+        var initSatVal = satisfactionValuesDict[this.model.get('intention').getInitialSatValue()].chartVal;
 
         // Add datapoints to graph for each userfunction/uservalue pair
-        var funcSegments = this.intention.dynamicFunction.getFuncSegmentIterable();
+        var funcSegments = this.model.get('intention').getFuncSegments();
+        console.log(funcSegments);
```
leshghi: Remove console log

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -717,10 +772,11 @@ var ElementInspector = Backbone.View.extend({
      * constraints from previously stored.
      */
     addConstraint: function(event) {
-
+
```
leshghi: This function needs to be renamed because it is not adding a constraint, it is adding a function segment. Same goes for the class name that calls this function

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -744,24 +800,25 @@ var ElementInspector = Backbone.View.extend({
...
     },
 
-
     /**
      * Toggles the display for the user defined function's
      * repeat feature.
      * This function is called on click for #constraint-repeat.
      */
     repeatConstraintControl: function(e){
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -807,12 +871,14 @@ var ElementInspector = Backbone.View.extend({
      * This function is called on change for #repeat-end2.
      */
     selectNumRepeatValues: function(event){
+
```
leshghi: Remove extra lines

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -933,35 +1011,41 @@ var ElementInspector = Backbone.View.extend({
...
+        
     },
 
     /**
      * Makes corresponding changes for the cell attributes, according to the values in the
      * inspector. This function is always called alongside with updateChart
      * and updateChartUserDefined.
      */
-    updateCell: function(event) {        
+    updateCell: function(event) {
```
leshghi: Remove event

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -933,35 +1011,41 @@ var ElementInspector = Backbone.View.extend({
...
      * and updateChartUserDefined.
      */
-    updateCell: function(event) {        
+    updateCell: function(event) {   
+            
         IntentionColoring.refresh();
         changeFont(current_font, paper);
-        var funcType = this.intention.dynamicFunction.stringDynVis;
-        var initSatVal = this.intention.getInitialSatValue();
+        if (this.model.get('intention').get('evolvingFunction') != null) {
```
leshghi: Can move logic around and inline variables in this function to make it much cleaner

leshghi (commented):
Okay wow amazing job! I know this looks like a lot of comments but a ton of them are minor things like removing extra lines, removing variables you only use once, etc and other cleanup stuff. I ignored the chart stuff, and gave some pointers for constraints, but I would check in again with Alicia about intended behavior and such for that.  I also probably missed a lot of stuff, but if you see me making a suggestion and then you think that suggestion can be implemented somewhere else as well (such as removing lines, inlining variables) please go for it. Also - I am not at all the authority on this so please please please feel free to push back on my comments if you disagree/I am wrong about something. That is what the reply function is for! Anyway - these are my initial thoughts. And again, amazing job guys this is a real heavy lifter part of the code and I know it has been a ton to take on.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: I think it might be the same? If it is the same, should we delete this function and just call initialValue directly?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: It was defined this way previously, @amgrubb do you know if this is correct?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
leshghi: I think so, if its the same then just calling that directly will be much more efficient and clean

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: I believe we would have to use slice() if we wanted to index it using -1

leaf-ui/rappid-extensions/BackboneModelsOther.js
i-m-foster: You want us to delete the whole file???

leaf-ui/rappid-extensions/ElementInspector.js
i-m-foster: Do you know what it means/ what it's used for?

leaf-ui/rappid-extensions/ElementInspector.js
i-m-foster: It shouldn't be possible to have a null evolving function here because this is only called for user defined functions. Also, where it is called is already inside an `if statement` for a null evolving function.

leaf-ui/rappid-extensions/BackboneModelsIntention.js
leshghi: Hmm okay tragic yeah I just looked into it and you can do like `var [lastValue] = yourArray.slice(-1)` but idk if thats worth it totally up to your preference

leaf-ui/rappid-extensions/ElementInspector.js
leshghi: Awesome - could you maybe write a comment documenting that somewhere in this function?

leaf-ui/rappid-extensions/ElementInspector.js
leshghi: Yeah, 13 is just the code that corresponds to the enter key, and its just so that when you press enter, the code knows to update the name (I believe)

leaf-ui/rappid-extensions/BackboneModelsOther.js
leshghi: No no just remove it from this PR

leaf-ui/rappid-extensions/ElementInspector.js
i-m-foster: Yup, at least one function segment should be in the list at this point. I'll add a comment

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: We are changing these defaults to fully satisfied for increasing and fully denied for decreasing

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: This function will need to be updated later, we are leaving it like this for now

leaf-ui/rappid-extensions/ElementInspector.js
i-m-foster: I think that we might need it because `initSatValueChanged()` calls `updateHTML()` which calls `validityCheck()` which actually requires event...

leshghi: This is actually maybe a much bigger issue that I should have caught way sooner but none of the class methods should actually touch the html...All html handling stuff should be done in the view, and then the model can contain model logic

leshghi: Basically anytime you need values from the html you should be getting those in the view and then passing them into a function in the model if you need to check validity/logic/etc

leshghi: This might be a stretch but thoughts on putting multiple function calls in your events? I know @YesugenB figured out how to do it...I am noticing that updateHTML() is usually called at the end of functions, and validityCheck() is then called in the first line of updateHTML(). Basically having a chain that goes "initSatValueChanged(), validityCheck(), updateHTML()" that way you don't have to pass the event parameter through 3 functions

leaf-ui/js/onFunctions.js
```diff
@@ -734,21 +734,12 @@ $('#colorblind-mode-isOn').on('click', function(){ //turns off colorblind mode
...
-    model.intentions.push(intention);
-
-    // create intention evaluation object
-    var intentionEval = new UserEvaluation(intention.nodeID, '0', '(no value)');
-    analysisRequest.userAssignmentsList.push(intentionEval);
-
-    cell.attributes.nodeID = intention.nodeID;
-
+    // create intention evaluation object and add it to userEvaluationList    
+    graph.get('userEvaluationList').push(new UserEvaluationBBM({intentionID: newIntentionBBM.cid, absTP: '0'}));
```
amgrubb: Sorry, can you clarify why this is happening? I thought we want to start with an empty list?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,339 @@
 
 /** 
  * Backbone Model of An Atomic Function Segment
+ * 
+ * @param {String} type
+ * Atomic function types
```
amgrubb, line 35: Add the types. is it `I,C,D,R' or longer names?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -7,61 +8,339 @@
...
+ * @param {String} refEvidencePair
+ * Corresponds to....
```
amgrubb: TODO, fill in the blank. ;)

leaf-ui/rappid-extensions/BackboneModelsIntention.js
amgrubb: Agreed that we need more comments here. Also, there is a bug, because if I change the value for a 'C' function, then it also resets the evolving function. Perhaps, we are missing an `else`?

amgrubb (changes requested):
Will add more changes later, once User Evaluations List is moved.

leaf-ui/rappid-extensions/ElementInspector.js
YesugenB: I changed events to this 
Screen Shot 2021-06-30 at 12 32 25 PM
and removed the event param from the initSatValueChanged and updateHTML (and commented out the function call for `updateHTML()` at initSatValueChanged and `validityCheck()` for updateHTML). 

Is this what you meant? I'm a little confused.

leaf-ui/js/onFunctions.js
YesugenB: We removed this, thank you!

leaf-ui/rappid-extensions/ElementInspector.js
YesugenB: We tried to implement this and got it working but it created other issues since we had to hide the other elements and we think it would be complicated to fix the issues.

leaf-ui/rappid-extensions/ElementInspector.js
leshghi: Yes - for the most part. I guess I was thinking if the first two functions only use the event param to pass it on to the last one, you could put all 3 in the backbone events section and pass "e" directly to only the last function if that makes sense

leaf-ui/rappid-extensions/ElementInspector.js
YesugenB: Nvm we got it!!!

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: It's all of the functions types

leaf-ui/rappid-extensions/ElementInspector.js
YesugenB: There are a few other functions that call updateHTML() and need to call validityCheck(). Should we do the same for those 2 functions?

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -236,18 +236,21 @@ var ElementInspector = Backbone.View.extend({
...
-        this.$('#segment-add').hide();
-        this.$('#constraint-repeat').hide();
-        this.$('#constraint-restart').hide();
         // Set correct dropdown options for function type based on initial satisfaction value
         $('option').show(); // Clear the previous selection
         if (this.intention.get('initialValue') == '(no value)'){
             // Hide all of the function options except for Stochastic is initial satisfaction value is '(no value)'
             $('option.B').hide(); 
+            this.$('#markedValue').hide();
+            this.$('#user-constraints').hide();
```
leshghi, line 244: I am pretty sure the user constraints div holds the segment add/constraint stuff, so I think you can just show/hide the user-constraints since that should also include the following 3 items

leaf-ui/js/onFunctions.js
```diff
@@ -771,9 +747,14 @@ graph.on("add", function(cell) {
...
     } else if (cell instanceof joint.shapes.basic.Intention){
-		cell.set('intention', new IntentionBBM({}));
-		cell.attr('.funcvalue/text', ' ');
-
+        // Creates an instance of a IntentionBBM from the cell
+        var newIntentionBBM = new IntentionBBM({})
+        cell.set('intention', newIntentionBBM);
+        cell.attr('.satvalue/text', '');
+        cell.attr('.funcvalue/text', ' ');
+        console.log(newIntentionBBM);
```
leshghi: Remove console log

leaf-ui/js/onFunctions.js
```diff
@@ -793,7 +774,7 @@ graph.on("add", function(cell) {
 
 // Auto-save the cookie whenever the graph is changed.
 graph.on("change", function(){
-	var graphtext = JSON.stringify(graph.toJSON());
+	var graphtext = graph.toJSON();
```
leshghi, line 777: Is removing the stringify the best solution here? I have no clue but know you guys were dealing with a circular dependency issue which feels like something larger that should be figured out/solved rather than avoided. I don't know what your final conclusion was on it though so just lmk!

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
  */
 
+/** 
+ * Backbone Model of UserEvaluations 
+ */ 
+ var UserEvaluationBBM = Backbone.Model.extend({  
+    initialize: function(options){  
+        _.extend({}, this.defaults, options)  
+        // TODO: Is absTP a letter or a number? It may need to be renamed.
+        this.absTP = options.absTP; // Integer Value
```
leshghi: Do we want to default this to 0? I feel like we could

leaf-ui/js/onFunctions.js
```diff
@@ -771,9 +747,14 @@ graph.on("add", function(cell) {
...
-		cell.attr('.funcvalue/text', ' ');
-
+        // Creates an instance of a IntentionBBM from the cell
+        var newIntentionBBM = new IntentionBBM({})
+        cell.set('intention', newIntentionBBM);
+        cell.attr('.satvalue/text', '');
+        cell.attr('.funcvalue/text', ' ');
+        console.log(newIntentionBBM);
+        // create intention evaluation object and add it to userEvaluationList 
+        newIntentionBBM.get('userEvaluationList').push(new UserEvaluationBBM({absTP: 0, assignedEvidencePair: '(no value)'}));
```
leshghi: If you do default absTP to 0 then you can just create a new UserEvaluationBBM with the default parameters

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+ * @param {String} startTP
+ * Start time point (char) 0,A,B,C
+ * @param {Integer} startAT
+ * Assigned/Absolute Time - Integer time value
+ * 
  */
 var FunctionSegmentBBM = Backbone.Model.extend({
+
     initialize: function (options) {
+        _.extend({}, this.defaults, options)
```
leshghi, line 47: This should only be done if you actually have a defaults section. Here at least I think startAT could definitely be a default

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+     * does not return anything.
+     */
+    getNthRefEvidencePair: function(n) { 
+        if (this.get('functionSegList') != null) {
+            var len = this.get('functionSegList').length; 
+            var funcSegList = this.get('functionSegList');
+            if (len > 1) {    
+                return funcSegList[len - n].get('refEvidencePair'); 
+            }   
+            else { // If the len is 1, can only return that FunctionSegmentBBM's refEvidencePair
```
leshghi, line 111: I don't know the use case but if you ask for the 10th item and you get the 1st I feel like that could be confusing potentially. Maybe it would make more sense to return null in that case?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+        _.extend({}, this.defaults, options) 
+    }, 
+
+    defaults: function(){ 
+        return {
+            nodeName: 'untitled',
+            nodeActorID: null,                     // Assigned on release operation.
+            nodeType: null,
+            evolvingFunction: null, 
+            // initialValue: '(no value)',
```
leshghi: Remove

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
-        nodeType: null,
-        nodeActorID: null,                     // Assigned on release operation.
-        evolvingFunction: null, 
-        initialValue: '(no value)'
-    },
 
     /**
-     * @returns Array of FunctionSegmentBBMs, or null if it is not an evolvingFunction
+     * Allows you to find the UserEvaluationBBM with the intentionID and absTP 
+     * TODO - we may need to edit this later if we remove intentionID
```
leshghi: Remove this todo and the cid param

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+     * 
+     * Sets the refEvidence pair based on the chosen initValue
+     * 
+     * @param {String} initValue 
+     */
+    changeInitialSatValue: function(initValue) {
+        // Set the the first element of userEvaluationList to the initValue
+        this.getUserEvaluationBBM(0).set('assignedEvidencePair', initValue);;
+        // this.set('initialValue', initValue);
+        var funcSegList = this.getFuncSegments();
```
leshghi: Can you put this inside the if statement since it won't return a valid value if the evolving function is null anyway

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+     * Sets the initial satisfaction value for this Intention to '(no value)'
+     */
+    removeInitialSatValue: function() {
+        this.changeInitialSatValue('(no value)');
+    },    
+ 
+    /**
+     * Resets evolving functions
+     */
+    // TODO: this function will need to be updated
```
leshghi, line 240: Does this still need to be updated?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+     * Clears all FuncSegments for the Intention's
+     * EvolvingFunction and adds new FuncSegments according to the current
+     * function type.
+     * Called when user changes the function type in Element Inspector 
+     * @param {String} funcType
+     */
+    setEvolvingFunction: function(funcType) {
+        this.set('evolvingFunction', new EvolvingFunctionBBM({type: funcType, functionSegList: []}));
+        var initValue = this.getUserEvaluationBBM(0).get('assignedEvidencePair');
+        // var initValue = this.get('initialValue');
```
leshghi: remove

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+
+
+        // Creates the correct FunctionSegmentBBM(s) for the selected function type
+        if (funcType == 'C' || funcType == 'R' || funcType == 'I' || funcType == 'D' || funcType == 'UD') {
```
leshghi, line 259: This would be a great case for a [switch](https://www.w3schools.com/js/js_switch.asp) I think

leaf-ui/rappid-extensions/BackboneModelsIntention.js
```diff
@@ -1,67 +1,380 @@
...
+            } else if (funcType == 'MN') {
+                // Decrease and Constant
+                var seg1 =  new FunctionSegmentBBM({type: 'D', refEvidencePair: null, startTP: '0', startAT: 0}); 
+                var seg2 =  new FunctionSegmentBBM({type: 'C', refEvidencePair: null, startTP: 'A', startAT: myNull}); 
+            } else if (funcType == 'SD') {
+                // Constant and Constant
+                var seg1 =  new FunctionSegmentBBM({type: 'C', refEvidencePair: '0011', startTP: '0', startAT: 0}); 
+                var seg2 =  new FunctionSegmentBBM({type: 'C', refEvidencePair: '1100', startTP: 'A', startAT: myNull}); 
+                this.getUserEvaluationBBM(0).set('assignedEvidencePair', '0011');
+                // this.set('initialValue', '0011');
```
leshghi: remove all outdated initial value comments in this function

leaf-ui/userguides/legend.html
```diff
@@ -98,11 +98,11 @@ <h3> Legend</h3>
             </tr>
             <tr>
                 <td> MP </td>
-                <td> Montonic Positive </td>
+                <td> Monotonic Positive </td>
```
leshghi, line 101: love

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -162,24 +162,21 @@ var ElementInspector = Backbone.View.extend({
...
      */
     render: function() {
-        /**
-        this.cell = cell; // Save the clicked node's backbone model
-
-        // Save the Intention object from the global model variable to
-        // this.intention
-        this.intention = model.getIntentionByID(this.cell.attributes.nodeID);
+        // Saves this.model.get('intention) as a local variable to access it more easily
+        this.intention = this.model.get('intention');
```
leshghi: I would do this in an initialize function

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -191,76 +188,79 @@ var ElementInspector = Backbone.View.extend({
         this.userConstraintsHTML = $("#new-user-constraints").last().clone();
 
         // Load initial value and node name
-        this.$('.cell-attrs-text').val(this.intention.nodeName);
-        this.$('#init-sat-value').val(satisfactionValuesDict[this.intention.getInitialSatValue()].name);
+        this.$('.cell-attrs-text').val(this.intention.get('nodeName'));
```
leshghi, line 190: This would be something you could set directly in the script if you wanted

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -476,12 +476,15 @@ var ElementInspector = Backbone.View.extend({
     validityCheck: function(event) {
         var functionType = this.$('.function-type').val();
         var initValue = this.$('#init-sat-value').val();
-
+        // Quick fix b/c functionType kept being reassigned null
```
leshghi: Maybe put a TODO for this?

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -93,8 +93,8 @@ var ElementInspector = Backbone.View.extend({
                     '<option value=D class="B"> Decrease </option>',
                     '<option value=RC class="B"> Stochastic-Constant </option>',
                     '<option value=CR class="B"> Constant-Stochastic </option>',
-                    '<option value=MP class="B"> Montonic Positive </option>',
-                    '<option value=MN class="B"> Montonic Negative </option>',
+                    '<option value=MP class="B"> Monotonic Positive </option>',
```
leshghi, line 98: Incredible

leshghi (approved):
This looks incredible guys! The structure is all there and it is working - awesome job. I left some minor comments and I think there is definitely some future cleanup/re-organizing we could do to try to somewhat simplify the view/move even more logic to the model but that can be done in the future. I'd love to get this merged in sooner rather than later just cause it is really a big part of the tool and now that it is working it will help everyone to be able to access it. There are still a few bugs that I noticed, which you can either deal with now or make issues for and deal with later after merging in this branch - totally up to you (I might even recommend the latter, as long as you make sure to make Issues and reference them in your PR description):
1. When you select an init value the (selected) "No Function" does not show up 
Screen Shot 2021-07-01 at 10 01 02 PM
2. I don't think the logic of options here is quite right
Screen Shot 2021-07-01 at 10 02 07 PM
3. You can only select the last time point for repeats
Screen Shot 2021-07-01 at 10 03 03 PM
4. I got an error when I clicked out of a UD Intention and then back in
Screen Shot 2021-07-01 at 10 12 34 PM

leaf-ui/js/onFunctions.js
i-m-foster: Catherine says that she thought JSON.stringify() is basically the same thing as .toJSON() so we only need one or the other not both at the same time. But we don't know too much too much about JSON...

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: Okay, I changed the default. I believe that absTP is an integer, so should we change the name to absAT?

leaf-ui/rappid-extensions/BackboneModelsIntention.js
i-m-foster: This function is only used to return either the last or second to last marked value. Which is then used to decide what satisfaction values should be displayed, so I think it should be okay.

YesugenB (approved):

amgrubb (approved):
Excellent work! There are definitely still some bugs that we will need to fix in the next sprint. I think this is good to merge, which will allow us to work on dependencies in other branches.

PR #463, opened 2021-07-01
Created Debug button to call graph when needed

Domily11:


amgrubb (approved):
I removed the extra code.

PR #473, opened 2021-07-02
Fixes ElementInspector.js bugs 

### LINKED ISSUE #466: Check logic of UD functions in ElementInspector.js
image

### LINKED ISSUE #467: Fix repeat time points for ElementInspector.js
You can only select the last time point for repeats

image

### LINKED ISSUE #468: Error when you click out of a UD intention and then back in
image

i-m-foster:
Closes #466 , #468 , #467

### DIFF
--- a/leaf-ui/rappid-extensions/ElementInspector.js
+++ b/leaf-ui/rappid-extensions/ElementInspector.js
@@ -200,7 +200,7 @@ var ElementInspector = Backbone.View.extend({
         
         // Turn off repeating by default
         this.repeatOptionsDisplay = false;
-        // TODO what does this do?
+        // Turn off display for repeat related elements and values
         this.setRepeatConstraintMode("TurnOff");
 
         // Load initial value for function type in the html select element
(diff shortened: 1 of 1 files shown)


leshghi (approved):
This looks good to me - I notice that the begin and end for repeats still reverts back to "Begin" and "End" when you switch away and back again. If you can fix that or make an Issue for it you should be set. Everything else seems and looks awesome. #466 does seem to be fixed somehow - incredible!

YesugenB (approved):

leshghi:
> This looks good to me - I notice that the begin and end for repeats still reverts back to "Begin" and "End" when you switch away and back again. If you can fix that or make an Issue for it you should be set. Everything else seems and looks awesome. #466 does seem to be fixed somehow - incredible!

Great job with the updates! I have found another bug (so sorry) with the repeat count and absolute length params - They only seem to save if you use the arrows to click them, not if you type in them. I am not sure why that is but I would guess it has something to do with different events being fired maybe? Also - would be cool if for repeats if you choose "B" as the starting point for example you could then not choose "C" from the ending point since the length between is less than two. That said this looks awesome for the bugs it was intended to fix and I think you should squash and merge and then just make issues to work on in a future PR for the two things I mentioned.

PR #486, opened 2021-07-08
Global Variable Removable

leshghi:
Removed as many global variables as I could in a comprehensive first round. This also involved some re-arranging of code. Definitely more to do but this felt like a good start and natural stopping point to check in. A lot of changes so definitely needs testing because I am sure there are a lot of bugs. Would appreciate a quick look over - happy to undo anything or do more, just let me know. Didn't or minimally touched backendComm, errorDetection, or loadSaveFunctions to try to avoid merge conflicts with @amgrubb and @Domily11. Also avoided analysis.js.

Happy to do more cleaning up/commenting/re-arranging of code but tried to keep in as in scope with removing global variables as possible while also doing some light cleanup/small bug fixes.

Also fixes UserEvaluationsList collection bug.


leaf-ui/js/onFunctions.js
```diff
@@ -505,16 +337,13 @@ paper.on('blank:pointerdown', function(evt, x, y) {
...
  */
 paper.on({
-    'cell:pointerdown': function(cellView, evt) {
-        var interact = true;
-        if(mode == "Analysis"){
-            interact = false;
-        }
-
+    // Note: cellView must remain in pointerdown and pointerup functions 
+    // In order for
```
leshghi, line 341: Sorry, didn't finish sentence here will remove in next commit

leaf-ui/js/onFunctions.js
```diff
@@ -632,104 +582,31 @@ function createHalo(cellView){
...
 		cellView.highlight();
     });
 
+    // Remove halo when cell is unhighlighted
+    halo.options.cellView.on('removeHalo', function(){
+        halo.remove();
+    });
+
     halo.render();
     return halo;
```
leshghi, line 596: Also will remove this return as it is no longer needed (comments for myself as well)

leaf-ui/js/displayAnalysis.js
```diff
@@ -133,58 +129,38 @@ function adjustSliderWidth(maxValue){
...
+    // analysisResult.selectedTimePoint = sliderValue;
 
-    var value = sliderValue;
-    $('#sliderValue').text(value);
-    sliderObject.sliderValueElement.innerHTML = value + "|" + currentAnalysis.timePointPath[value];
-    // Update the analysisRequest current state.
-    analysisRequest.currentState = sliderObject.sliderValueElement.innerHTML;
-    currentAnalysis.elementList.forEach(element => 
-        updateNodeValues(element.id, element.status[value]));
+    // var value = sliderValue;
```
amgrubb, line 137: Yes, I think it is a good protocol to comment out the code that needs to be updated with the new backbone models, but leave the functions call there with a TODO. :)

leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -69,6 +69,14 @@ var ElementInspector = Backbone.View.extend({
         this.listenTo(this, 'change: intention', this.initSatValueChanged); 
         // Saves this.model.get('intention) as a local variable to access it more easily
         this.intention = this.model.get('intention');
+        this.satValueDict = {
```
amgrubb, line 72: Is this second dictionary required?

amgrubb (approved):
I agree this is an extensive first pass and much testing will be required.
So far so good. 👍

leaf-ui/rappid-extensions/ElementInspector.js
leshghi: Not if we replace the values in the option tags with their binary string equivalents, but I would need to go through further to determine if that would make things easier/more confusing. Happy to do that here or in a future PR

PR #488, opened 2021-07-09
Cleanup Part 2

leshghi:
1. Removed extra sat value dicts
2. Removed AnalysisResult JS class and added the params to ResultBBM
3. Cleaned up switchToAnalysisMode and switchToModellingMode

In draft form because still a WIP but I want @YesugenB and @i-m-foster to take a look at my ElementInspector.js changes to help make sure I didn't break anything/create huge issues for them.


leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -339,7 +353,7 @@ var ElementInspector = Backbone.View.extend({
         funcSegments.forEach(
             element => {
```
i-m-foster: We could change `element ` to `functionSegment ` here

i-m-foster (approved):
It looks good to me! I don't think it should cause any major issues for us

leaf-ui/rappid-extensions/ElementInspector.js
leshghi: Nice will do in my next commit

YesugenB (approved):

amgrubb (approved):
Wow, that's a lot of cleanup. Do you have a formatting tool that corrects the spaces? If so, what do you use/recommend?

leshghi:
> Wow, that's a lot of cleanup. Do you have a formatting tool that corrects the spaces? If so, what do you use/recommend?

VSCode does it for you! If you right click and select format document from the menu it will do the whole file. You can also highlight just a portion of code and then choose format selection.

PR #490, opened 2021-07-12
Fixed #456

### LINKED ISSUE #456: Potential duplicate placeholder names for Actor
If you make Actor_0, Actor_1, and then delete Actor_0 and make a new actor then you will have two Actor_1s

Domily11:
The function now reads the number of the last actor so that even when previous actors are deleted the function still keeps adding up for the default actor name

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -236,7 +236,16 @@ graph.on("add", function (cell) {
         createdInstance = createdInstance.filter(view => view.model instanceof joint.shapes.basic.Actor);
 
         // Create placeholder name based on the number of instances
-        var name = cell.attr('.name/text') + "_" + (createdInstance.length - 1);
+        if (createdInstance.length >= 2){
+            var lastactor = createdInstance[createdInstance.length - 2].model.attr('.name/text');
+            var num = Number.parseInt(lastactor.charAt(lastactor.length-1))+1
+            var name = cell.attr('.name/text') + "_" + num;
+
+        }else{
+            var name = cell.attr('.name/text') + "_0";
+        }
+
+        
         cell.set('actor', new ActorBBM({ actorName: name }));
         cell.attr(".name/text", name);
 
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -236,16 +236,14 @@ graph.on("add", function (cell) {
         createdInstance = createdInstance.filter(view => view.model instanceof joint.shapes.basic.Actor);
 
         // Create placeholder name based on the number of instances
+        var name;
         if (createdInstance.length >= 2){
             var lastactor = createdInstance[createdInstance.length - 2].model.attr('.name/text');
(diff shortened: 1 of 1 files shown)


i-m-foster (approved):

leaf-ui/js/onFunctions.js
```diff
@@ -236,7 +236,16 @@ graph.on("add", function (cell) {
         createdInstance = createdInstance.filter(view => view.model instanceof joint.shapes.basic.Actor);
 
         // Create placeholder name based on the number of instances
-        var name = cell.attr('.name/text') + "_" + (createdInstance.length - 1);
+        if (createdInstance.length >= 2){
+            var lastactor = createdInstance[createdInstance.length - 2].model.attr('.name/text');
+            var num = Number.parseInt(lastactor.charAt(lastactor.length-1))+1
```
YesugenB: Do you think we could inline the `num`? Since it's only used once here. Would need` ( )` around it I tho I think

leaf-ui/js/onFunctions.js
```diff
@@ -236,7 +236,16 @@ graph.on("add", function (cell) {
...
+            var name = cell.attr('.name/text') + "_" + num;
+
+        }else{
+            var name = cell.attr('.name/text') + "_0";
+        }
+
+
```
YesugenB: Remove extra line

leaf-ui/js/onFunctions.js
```diff
@@ -236,7 +236,16 @@ graph.on("add", function (cell) {
         createdInstance = createdInstance.filter(view => view.model instanceof joint.shapes.basic.Actor);
 
         // Create placeholder name based on the number of instances
-        var name = cell.attr('.name/text') + "_" + (createdInstance.length - 1);
+        if (createdInstance.length >= 2){
+            var lastactor = createdInstance[createdInstance.length - 2].model.attr('.name/text');
+            var num = Number.parseInt(lastactor.charAt(lastactor.length-1))+1
+            var name = cell.attr('.name/text') + "_" + num;
+
```
YesugenB: Can remove the extra line here and maybe add a space before and after the `else` so that it's uniform with the previous `else` statements

YesugenB (approved):
Just some small comments! Looks really good though :>>

leaf-ui/js/onFunctions.js
```diff
@@ -236,7 +236,16 @@ graph.on("add", function (cell) {
         createdInstance = createdInstance.filter(view => view.model instanceof joint.shapes.basic.Actor);
 
         // Create placeholder name based on the number of instances
-        var name = cell.attr('.name/text') + "_" + (createdInstance.length - 1);
+        if (createdInstance.length >= 2){
```
amgrubb, line 240: var name should be defined outside the if statement.

amgrubb (changes requested):
Just a small change.

leaf-ui/js/onFunctions.js
Domily11: Very valid, I'll change it

amgrubb (approved):
Awesome, great.

PR #494, opened 2021-07-13
Result (Configuration) Deletion Fix 

### LINKED ISSUE #465: Analysis Parameters need to switch to new Configuration
Currently in ConfigInspector.js, when you delete a selected configuration, the selection switches over but the analysis parameters do not. Needs to be fixed so that the analysis parameters switch to the newly selected config as well. Additionally, please add a catch for when there is only 1 config and it gets deleted, so that a new default config is automatically made and added back

angelac2023:
If a result is deleted, then the user no longer has to click the updated selected result in order to see its analysis parameters. The analysis parameters update immediately following the result deletion.

Closes #465


amgrubb:
Looks good, waiting on another team member to review.

YesugenB (approved):
Looks good! Good stuff. Just gotta fix the little conflict

amgrubb (approved):
Yup, agreed, just fix the conflict before merging.

PR #497, opened 2021-07-16
Renamed absTime in LinkBB to absTP

### LINKED ISSUE #493: Rename LinkBBM absTime to absTP
LinkBBM absTime should be defaulted to null.
Be sure to follow that change through in LinkInspector.js, AssignmentsTable.js, and IntermediateValues.js.

Manarnz:
This fixes the issue #493


leaf-ui/rappid-extensions/AssignmentsTable.js
```diff
@@ -383,7 +383,7 @@ var LinkRelationshipView = Backbone.View.extend({
         '<td><%= linkType %><%if (evolving){%> | <%}%> <%=postType %></td>',
         '<td class= "link-source"></td>',
         '<td class= "link-dest"></td>',
-        '<td><input class="linkAbsRelation" type="number" name="sth" value= <%- absTime %> > </td>',
+        '<td><input class="linkAbsRelation" type="number" name="sth" value= <%- absTP %> > </td>',
```
YesugenB: Wondering if we can remove the `name="sth"` as I don't think we really need it? Not sure though, what does it do 🧐

YesugenB (approved):
Looks good, just had a small question/suggestion that's not very related to this. 

Btw in AssignmentsTable.js line 43 Epoch is misspelled as Epcoch.

leaf-ui/rappid-extensions/AssignmentsTable.js
Manarnz: I don't know what it does. I think maybe I should stick to the task? What do you think?

YesugenB: Yeah it's fine! No need to change then

i-m-foster (approved):

amgrubb:
@Manarnz  Do AssignmentsTable.js lines 400 and 425 need to be updated with myNull as well?

Manarnz:
> @Manarnz Do AssignmentsTable.js lines 400 and 425 need to be updated with myNull as well?

I have checked these two lines and yes they do need change. For line 400, I removed the if statement as the TODO comment suggested.

leaf-ui/rappid-extensions/AssignmentsTable.js
```diff
@@ -395,11 +395,7 @@ var LinkRelationshipView = Backbone.View.extend({
 
     render: function () {
         this.$el.html(_.template($(this.template).html())(this.model.toJSON()));
-        // TODO: Write statement to handle this case in script
-        // Or if absTime default changed to null just remove if statement
-        if (this.model.get('absTime') == -1) {
-            this.$('.linkAbsRelation').val('')
-        }
+        this.$('.linkAbsRelation').val('')
```
amgrubb: The `if` statement must remain because otherwise save values will not appear in the view. Switch the original if statement to `myNull` instead of `-1`.

amgrubb (changes requested):
We need to put the `if` statement back.

amgrubb:
Please address or resolve the conversations above, I'll review it after that. 😄

amgrubb:
- [x] There is still an issue, when you click `Unassign` for one relationship, it unassigns all of them.

amgrubb (approved):
Excellent work!

PR #504, opened 2021-07-21
Fixed #499

### LINKED ISSUE #499: Fix Paper Size in Analysis Model
image

Domily11:
Paper now expands when changed to analysis view

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -433,6 +433,9 @@ paper.on("link:options", function (cell) {
 
         // Remove model only elements 
         $('.model-only').css("display", "none");
+        $('.inspector').css("display", "none");
+        $('#paper').css("right", "0px");
+
 
         // Show extra tools for analysis mode
         $('.analysis-only').css("display", "");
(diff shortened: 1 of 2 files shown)


amgrubb:
@Domily11 

- [x] There is a bug with the relationship inspector, I think you need to make another update. When navigating back from the analysis view the relationship inspector doesn't look correct.

Domily11:
> @Domily11
> 
> * [x]  There is a bug with the relationship inspector, I think you need to make another update. When navigating back from the analysis view the relationship inspector doesn't look correct.

Yep, it now works correctly I believe

leaf-ui/js/onFunctions.js
```diff
@@ -433,6 +433,9 @@ paper.on("link:options", function (cell) {
 
         // Remove model only elements 
         $('.model-only').css("display", "none");
+        $('.inspector').css("display", "none");
+        $('#paper').css("right", "0px");
+
```
Manarnz, line 438: Maybe delete this additional line (I'm trying to be like Ali ahah).

Manarnz (commented):
You did a great job Cat! Nice solution to the bug.

amgrubb (approved):
Looks great! Please `Squash and merge`.

PR #506, opened 2021-07-21
Fixed Issue #505

Domily11:
The titles now show up for LinkInspector


amgrubb:
@Domily11 Instead, can you merge this into `ACK-paper-expansion` and see if it fixes my issue with #504.

PR #536, opened 2021-08-10
Fixed chart issue

Manarnz:
Before, when changing abs time, the graph would update even though it does not need to.


YesugenB (dismissed):

i-m-foster (changes requested):
The chart no longer updates for non-UD functions. Also, previously in UD functions whenever you updated the function type or satisfaction value the chart changed, but now it only changes when you click add.

i-m-foster (approved):

amgrubb (approved):
I like this solution. 👍

PR #562, opened 2021-08-20
Formatted documents

YesugenB:


amgrubb (approved):
Looks good, wow it changes so much!

PR #566, opened 2021-09-09
Resolved remaining CSS bugs

### LINKED ISSUE #326: General CSS bugs
Intending this to be a place to compile CSS bugs to fix at a future date.

- [x]  I can scroll down on the webpage because the analysis inspector panel is too long. (Sometimes I can scroll side-to-side, too, especially in a smaller window.)
- [x]  The watermark overlaps the config scroll box/the actors when the webpage gets shorter.
- [x]  We need to update the watermark with Smith info
- [x]  Could we auto-scroll to where new results are added? Or else add a less shadow-y line at the bottom of the scroll box so the user can see it's scrolling.
- [x]  Larger EVO slider overlaps the View Assignments List button

krhablutzel:
- changed stencil scroll box to stop above the watermark/attribution box, so attribution doesn't overlap stencil elements
- investigated auto-scrolling to where new results are added; decided this would be overcomplicated
- verified that other listed bugs appear to be fixed

### DIFF
--- a/leaf-ui/css/style.css
+++ b/leaf-ui/css/style.css
@@ -119,9 +119,14 @@ body {
   bottom: 0;
   width: 200px;
   background-color: #f9f9f9;
-  box-shadow: 2px 0px 16px 0px rgba(0,0,0,0.2);
+  
   /* Should be less than the modal pop-ups */ 
   z-index: 0;
+}
+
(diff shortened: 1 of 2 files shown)


krhablutzel:
(issue #326)

i-m-foster (approved):

PR #572, opened 2021-09-14
Config names no longer overlap

Vbeany:
Known bugs: Changing name of last config gives Request NaN for next config. Made separate issue #573

### DIFF
--- a/leaf-ui/rappid-extensions/ConfigInspector.js
+++ b/leaf-ui/rappid-extensions/ConfigInspector.js
@@ -347,7 +347,13 @@ var ConfigInspector = Backbone.View.extend({
 
     /** Create and add a new config model to the collection */
     addNewConfig: function () {
-        var configModel = new ConfigBBM({ name: "Request " + (this.collection.length + 1), results: new ResultCollection([]) })
+        var configModel; 
+        if (this.collection.length < 1){
+                configModel = new ConfigBBM({ name: "Result " + 1, results: new ResultCollection([]) });
+        } else { 
+            var nameIndex = parseInt(this.collection.at(this.collection.length - 1 ).get('name').split(' ').pop()) + 1;
+            configModel = new ConfigBBM({ name: (this.collection.at(this.collection.length - 1).get('name').split(' ')[0] + " " + nameIndex), results: new ResultCollection([]) })
+        }
         this.collection.add(configModel);
     },
 });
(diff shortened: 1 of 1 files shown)


amgrubb:
I changed the base to `sep21-bugs`.

i-m-foster (approved):
It looks good to me! I would add the bug number it fixes in a comment so it's easy to see and maybe also add the bug number to the linked issues section on the side.

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -347,7 +347,13 @@ var ConfigInspector = Backbone.View.extend({
 
     /** Create and add a new config model to the collection */
     addNewConfig: function () {
-        var configModel = new ConfigBBM({ name: "Request " + (this.collection.length + 1), results: new ResultCollection([]) })
+        var configModel; 
+        if (this.collection.length < 1){
+                configModel = new ConfigBBM({ name: "Result " + 1, results: new ResultCollection([]) });
```
krhablutzel: We want to call these "Request", not "Result"

leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -347,7 +347,13 @@ var ConfigInspector = Backbone.View.extend({
...
+        } else { 
+            var nameIndex = parseInt(this.collection.at(this.collection.length - 1 ).get('name').split(' ').pop()) + 1;
```
krhablutzel, line 354: When a request is renamed to something not containing a number, this replaces the last word in the name with `NaN`. This also returns odd results when a Request is named just a number or has an empty name. (Also, less importantly, if we rename a Request to a different number, we start counting from that number even if that isn't the total number of Requests ever generated - not sure whether we like this behavior.)

Screen Shot 2021-09-18 at 13 34 47

anishaj149 (changes requested):
(Worked with Kathleen so her comment contains necessary changes.) :)

krhablutzel (changes requested):
This works well when the user has not renamed Requests (other than the Result/Request typo). However, it generates some odd behavior when the user has renamed the Request (see comments below).

krhablutzel (approved):
Approving this since the noted changes are covered in separate issue #573

PR #583, opened 2021-09-18
Change color of disabled functions

### LINKED ISSUE #579: Init sat value box turns dark green
When you click away from an intention and then back the initial satisfaction value box turns green. So far, I've only seen it happen on UD functions
image

Vbeany:
Closes #579

### DIFF
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -247,11 +247,11 @@ body {
 }
 
 #init-sat-value:disabled {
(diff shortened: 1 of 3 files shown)


i-m-foster (changes requested):
Nice the color bug looks like its fixed! There's still a bug where the initial satisfaction value becomes disabled for UD functions once you leave an intention and go back. Could fix that too??

i-m-foster (approved):
looks good😈

Issue #551, opened 2021-08-13
When config is deleted, EVO slider is not updated to modeling mode.


Manarnz:
When deleting a config with results (while one of its results is selected) the EVO slider does not update to modelling slider.
![image](https://user-images.githubusercontent.com/83615524/129415102-2f6f7daf-18a4-4e26-8255-26cb102cdc34.png)
![image](https://user-images.githubusercontent.com/83615524/129415119-f051e73d-d14b-4509-a63a-976f1f208a0a.png)

Additionally, when deleting the last config and it has simulated path, the AnalysisInspector does not update. We can solve this issue by making it automatically create a new config if there aren't any.
![image](https://user-images.githubusercontent.com/83615524/129415136-7bc7ae19-f870-4654-ad39-d54103b6d8b2.png)

PR #584, opened 2021-09-19
Ytck evo bugs

### LINKED ISSUE #564: EVO overwrites the satisfaction labels so that they all become the same label
The satisfaction values vary amongst the nodes when the result first comes out, but if we switch to EVO state, the labels are changed to be completely the same.

image

image

YanningTan:
Fix #564 and #551

### DIFF
--- a/leaf-ui/rappid-extensions/ConfigInspector.js
+++ b/leaf-ui/rappid-extensions/ConfigInspector.js
@@ -185,6 +185,7 @@ var Config = Backbone.View.extend({
             }
         }
         this.model.destroy();
+        IntentionColoring.setColorMode("EVO", undefined);
     },
 
     /** If result is selected, update config **/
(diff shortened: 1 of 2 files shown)


leaf-ui/rappid-extensions/ConfigInspector.js
```diff
@@ -185,6 +185,8 @@ var Config = Backbone.View.extend({
             }
         }
         this.model.destroy();
+        IntentionColoring.setColorMode("EVO", undefined);
```
YesugenB: Is IntentionColoring still being used? I thought we were removing it from evoObjects

YesugenB:
Maybe make a separate issue for the part of #551 that is not yet completed?

YesugenB (approved):

PR #585, opened 2021-09-19
Fixed #571

### LINKED ISSUE #571: Intermediate Values Table at the bottom is cut off when viewing
Only for a couple of models: When viewing the Intermediate Values Table with lots of info and scrolling down to look the last row is cut off from view
(Screenshot from vendor model)
Cut off

Domily11:


leaf-ui/js/onFunctions.js
```diff
@@ -211,6 +211,7 @@ $('#btn-view-intermediate').on('click', function () {
     var intermediateValuesTable = new IntermediateValuesTable({ model: graph });
     $('#intermediate-table').append(intermediateValuesTable.el);
     intermediateValuesTable.render();
+    $('.intermT').height($('#paper').height() * 0.9);
```
YesugenB, line 214: Do you need this here if you have height set as 90% in the css? Or vice verse, do you need to set the height as 90% in the css if you have * 0.9 here? 

Venus tried commenting out each of these, and it seems to work with just one of these changes

YesugenB (approved):
Looks good! Just had one question

Issue #573, opened 2021-09-14
Changing name of last config causes naming issues


Vbeany:
If you change the name of the las config and then add a new config, the new config is named Request NaN. Similar to Issue #570

Issue #570, opened 2021-09-11
Adding new actor after renaming previous one causes Actor_Nan


YesugenB:
Here, I renamed `Actor_0` to`Metro`. Adding more actors after this causes the following actor names to become `Actor_Nan`
<img width="325" alt="Screen Shot 2021-09-11 at 2 30 00 PM" src="https://user-images.githubusercontent.com/59470488/132957775-b8fb22b6-137e-42ad-9d12-f73f3f35f2fd.png">

PR #587, opened 2021-09-28
Fixed naming issue with NaN 

Vbeany:
Closes issue #573 and #570


i-m-foster (approved):
The only potential issue I noticed is if you rename an actor with a number (like Actor_2), the numbers reset so you can end up with a duplicate name

Issue #592, opened 2021-10-05
Constraint does not update when intention is changed


Domily11:
When an intention changes from an evolving to a non-evolving function, the constraint associated with it does not automatically get removed

PR #597, opened 2021-10-08
Solved #592

Domily11:
Added a function in ElementInspector that destroyed associated constraint of intentions that were previously evolving functions


Domily11:
@amgrubb would destroying all the associated constraints like this work or would you like me to keep the constraint if the new function type is also an evolving function?

amgrubb:
@Domily11 We agreed that if you change the evolving function, it resets all the values associated with that intention in the Intermediate Values Table and the Absolute & Relative Assignments Table.

Domily11:
@i-m-foster Hey Irene, gonna need you to hold off this review for a bit, the bug apparently wasn't completely fixed ugh

i-m-foster (approved):
it looks good to me!

PR #598, opened 2021-10-13
Dragging actor under intentions, and removing mentions of configurations

krhablutzel:
Resolves issue 577 (replace mentions of Configuration with Request in the frontend), and makes progress towards resolving issue 345. Moving an actor under an intention will now add the intention to the actor - but only as long as the intention aligns with the the center of the actor. We still need to find a way to add the intention to the actor if it's not perfectly aligned with the actor's center.


YesugenB (approved):
Looks great!

i-m-foster (approved):
Looks good👍

PR #599, opened 2021-10-15
Code clean up and fixing small bugs

### LINKED ISSUE #462: Simplifying the code to switch back and forth from analysis and modelling views re: html
Right now we are accessing a whole lot of html elements in switchToModellingMode and switchToAnalysisMode in onFunctions.js and it can be a bit difficult to parse through. It would be great if we could have a common class/attribute shared by the toolbar elements that are supposed to be shown in modeling mode, and the same for analysis mode so that we could access them to show/hide in one go. (Similar to the technique implemented in switchToAnalysisMode() : `$('.model-clears').css("display", "none"); $('.analysis-clears').css("display", "");`).

### LINKED ISSUE #565: EVO Automatically Turned On
If we switch to the analysis mode, click ‘back’ button on Chrome, and then return to the BloomingLeaf, EVO would be automatically turned on.
WechatIMG287

### LINKED ISSUE #595: Enable no value in Element Inspector
Should be able to select no value in element inspector

i-m-foster:
Closes #595 #565 and #462

### DIFF
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -309,7 +309,7 @@ select > .select-placeholder {
   font-weight: 200;
 }
 
-.link-tools .tool-options{
+.link-tools{
    display: block;
 }
 .inspector-btn-small {
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -582,15 +582,6 @@ select > .select-placeholder {
   z-index: 3;
 }
 
-#modeText{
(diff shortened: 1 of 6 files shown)


Domily11 (changes requested):
So the problem with #565 is still there I think, the slider is turned off but the color still changed when you changed the initial sat value. I'm not sure if it's a my-computer issue tho. Also, there was another small bug that I found (#603) and was wondering if you guys want to take it over as well

Domily11 (approved):

PR #612, opened 2021-10-30
Implementing EVO into Nextstate 

### LINKED ISSUE #600: Implement EVO into next state and adding an Intention Filter to Next State

Vbeany:
Closes #600


amgrubb:
@Domily11 @krhablutzel Any chance you could review this sometime this weekend?

leaf-ui/js/object/evoObjects.js
```diff
@@ -597,26 +597,44 @@ class EVONextState {
...
-        var step = 0;
+        allSolutionArray = [];      
+        // Iterates over the hashmap allSolutions and combines all of the solutions into one array
+        for (var key in myInputJSObject.results.get('allSolutions')) {
+            // Adds every element (which are arrays) in the old array to the new array
+            myInputJSObject.results.get('allSolutions')[key].forEach(
+                solution => {
+                    allSolutionArray.push(solution);
+                })
+        }
```
Domily11, line 608: Add ;?

Domily11 (approved):
Everything looks great!

krhablutzel (commented):
What happened to the slider in EVO state mode? I thought we were supposed to be able to view the colors/states of each intention at a certain time point, and then use the slider to scroll through timepoints.

i-m-foster:
> What happened to the slider in EVO state mode? I thought we were supposed to be able to view the colors/states of each intention at a certain time point, and then use the slider to scroll through timepoints.

Next State is all of the possible solutions at one time point, so to see the different colors/states of each intention you switch to the solution you want to see instead of having a slider for all of the solutions. At least that's what i think we were supposed to do lol

krhablutzel:
> > What happened to the slider in EVO state mode? I thought we were supposed to be able to view the colors/states of each intention at a certain time point, and then use the slider to scroll through timepoints.
> 
> Next State is all of the possible solutions at one time point, so to see the different colors/states of each intention you switch to the solution you want to see instead of having a slider for all of the solutions. At least that's what i think we were supposed to do lol

Got it, that makes sense for Next States. However, the slider now appears to be missing in state mode on the main analysis page.

i-m-foster:
> > > What happened to the slider in EVO state mode? I thought we were supposed to be able to view the colors/states of each intention at a certain time point, and then use the slider to scroll through timepoints.
> > 
> > 
> > Next State is all of the possible solutions at one time point, so to see the different colors/states of each intention you switch to the solution you want to see instead of having a slider for all of the solutions. At least that's what i think we were supposed to do lol
> 
> Got it, that makes sense for Next States. However, the slider now appears to be missing in state mode on the main analysis page.

There's a bug where the slider disappears in EVO state mode if you have an actor, but that bug was there previously. If you have no actors does the slider in EVO state mode still disappear for you?

PR #619, opened 2021-11-19
Adding SWAL errors and fixing font

### LINKED ISSUE #593: Font changes when switch from analysis back to modelling

### LINKED ISSUE #603: Clear Evaluation Button gives an error when there is not function assigned to intention

### LINKED ISSUE #604: Disable opening Next State if the EVO slider is on % or time modes

YesugenB:
Closes #593,  #603 and #604

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -798,13 +798,19 @@ paper.on("link:options", function (cell) {
         for (let element of graph.getElements()) {
             var cell = element.findView(paper).model;
             var intention = cell.get('intention');
-            var initSatVal = intention.getUserEvaluationBBM(0).get('assignedEvidencePair');
-            var funcType = intention.get('evolvingFunction').get('type');
+            
+            if (intention != null) {
+                var initSatVal = intention.getUserEvaluationBBM(0).get('assignedEvidencePair');
+                if (intention.get('evolvingFunction') != null) {
+                    var funcType = intention.get('evolvingFunction').get('type');
+                }
+            }
 
             // If the initsatVal is not empty and if funcType empty
             if (intention != null && initSatVal != '(no value)' && funcType === 'NT') {
                 intention.removeInitialSatValue();
                 cell.attr(".satvalue/text", "");
+                $('#init-sat-value').val('(no value)');
             }
         }
         EVO.refresh(selectResult);
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -173,7 +173,7 @@ $('#btn-clear-flabel').on('click', function () {
         var cellView = element.findView(paper);
         var cell = cellView.model;
         var intention = cell.get('intention');
-        if (intention != null && intention.get('evolvingFunction').get('type') != 'NT') {
+        if (intention != null && intention.get('evolvingFunction') != null && intention.get('evolvingFunction').get('type') != 'NT') {
             intention.removeFunction();
             cell.attr(".funcvalue/text", "");
         }
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -535,16 +535,22 @@ paper.on("link:options", function (cell) {
             curRequest.set('action', 'allNextStates');
             curRequest.set('previousAnalysis', curResult);
 
-            // If the last time point is selected, error message shows that you can't open Next State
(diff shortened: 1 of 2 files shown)


leaf-ui/js/onFunctionsNextState.js
```diff
@@ -66,8 +72,14 @@ function fontUp(pPaper) {
  * @param {*} pPaper 
  */
 function fontDown(pPaper) {
-    var new_font = current_font - 1;
+    var elements = analysis.graph.getElements();
+    for (var i = 0; i < elements.length; i++) {
+        var cellView = elements[i].findView(pPaper);
```
krhablutzel, line 77: This is also true in develop, but font sizes don't change on intentions on some of the saved models, such as scheduler or half the intentions in WME. Perhaps an issue with the saved models rather than this code?

leaf-ui/js/onFunctions.js
```diff
@@ -794,13 +798,19 @@ paper.on("link:options", function (cell) {
...
+                if (intention.get('evolvingFunction') != null) {
+                    var funcType = intention.get('evolvingFunction').get('type');
+                }
+            }
 
             // If the initsatVal is not empty and if funcType empty
             if (intention != null && initSatVal != '(no value)' && funcType === 'NT') {
                 intention.removeInitialSatValue();
                 cell.attr(".satvalue/text", "");
+                $('#init-sat-value').val('(no value)');
```
krhablutzel, line 813: Heads up that this changes to read '(no value)' even if the user has selected a node with an evolving function type (whose initial sat value doesn't clear). However, the user can click the intention again to refresh the inspector to show the correct initial sat value, so it's probably not a big issue.

krhablutzel: To recreate, add a node with an initial satisfaction value but no function, then add another node with an initial sat value + an evolving function, then clear evaluation labels. (Evaluation labels are only cleared if there is at least one node available on which to clear labels.)

leaf-ui/js/onFunctions.js
```diff
@@ -794,13 +798,19 @@ paper.on("link:options", function (cell) {
         for (let element of graph.getElements()) {
             var cell = element.findView(paper).model;
             var intention = cell.get('intention');
-            var initSatVal = intention.getUserEvaluationBBM(0).get('assignedEvidencePair');
-            var funcType = intention.get('evolvingFunction').get('type');
+            
+            if (intention != null) {
+                var initSatVal = intention.getUserEvaluationBBM(0).get('assignedEvidencePair');
+                if (intention.get('evolvingFunction') != null) {
+                    var funcType = intention.get('evolvingFunction').get('type');
```
krhablutzel, line 805: I still get an error if I "Clear Evaluation Labels" on a node that used to have an evolving function, but has been changed back to "No Function". (You have to click off the node to assign the evolving function, then click back onto the node to return it to "No Function".)

krhablutzel (approved):
Looks great to me! I left some comments on a few small remaining bugs.

leaf-ui/js/onFunctions.js
YesugenB: Screen Shot 2021-12-03 at 5 34 09 PM
Was this the error you got?

leaf-ui/js/onFunctionsNextState.js
Vbeany: When you use the V2 models that problem does show up but using the V3 models all the font sizes do change. There is an issue with the V3 models where when you change the font sizes and then go into analysis mode and back to modeling mode the font sizes for some of the intentions reset to the default size.

leaf-ui/js/onFunctions.js
```diff
@@ -530,21 +530,25 @@ paper.on("link:options", function (cell) {
...
             curRequest.set('previousAnalysis', curResult);
 
-            // If the last time point is selected, error message shows that you can't open Next State
-            if ((curResult.get('timePointPath').length - 1) === curResult.get('selectedTimePoint')) {
-                swal("Error: Cannot explore next states with last time point selected.", "", "error");
-                $("body").removeClass("spinning"); // Remove spinner from page
-            } else {
-                backendSimulationRequest(curRequest);
+            if (EVO.sliderOption == '1' || EVO.sliderOption == '2') {
+                swal("Error: Cannot explore next states from percent or time EVO options.", "", "error");
```
Domily11, line 538: Not sure if you still want this message when Evo is turned on but the user hasn't clicked a Result yet

image

Domily11 (commented):
Everything else looks great except for that one small question I had!

PR #626, opened 2022-02-03
Added new option participates-in

Domily11:
Finished #625


krhablutzel (approved):
Thanks for adding this!

PR #627, opened 2022-02-03
Fixed #613

### LINKED ISSUE #613: If you have an intention selected, you can't clear the evaluation label after you clear dynamic label
This is because while clear dynamic label clear the function text in the intention, it does not clear the function-type in the element inspector. It should be fixed by rerendering/updating html for element inspector

sonorah:


amgrubb:
@sonorah Please add a description to your pull request.

i-m-foster (changes requested):
Looks good, except if you click on "Clear Dynamic Function Labels" multiple times you get an error

image

PR #633, opened 2022-02-10
Fixed #616

### LINKED ISSUE #616: On windows computers the BloomingLeaf interface has two scrollers
This issue appears in analysis mode and next state.
image

sonorah:
The buttons were too large and adding extra space to the window. We made it so that the user can scroll down the left panel instead

### DIFF
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -109,7 +109,7 @@ body {
   top: 40px;
   bottom: 0;
   width: 200px;
-
+  overflow-y: scroll;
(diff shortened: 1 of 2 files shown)


Vbeany:
I still have two scrollers when my window is small. 
image

Vbeany (changes requested):
There's only one scroller when the window is big but there are still two scrollers when the window is small.

Vbeany:
I no longer have two scrollers, looks good!

Vbeany (approved):
Only one scroller now. Looks good!

PR #642, opened 2022-03-10
Created New Function clearFunctionSegments to fix #622

### LINKED ISSUE #622: Element Inspector Doesn't Automatically update When Clear Dynamic Function Label is clicked
image

After the function label is cleared, the function is cleared as well, but the element inspector is not automatically updated to reflect that

image

sonorah:
Also @YesugenB and @i-m-foster  can you guys determine if we still need the removeFunctions if we didn't use it at all?
Fixed 613 (the error with clearing evaluation labels).

### DIFF
--- a/leaf-ui/rappid-extensions/ElementInspector.js
+++ b/leaf-ui/rappid-extensions/ElementInspector.js
@@ -68,6 +68,7 @@ var ElementInspector = Backbone.View.extend({
         this.listenTo(this, 'change:intention', this.initSatValueChanged);
         // Saves this.model.get('intention) as a local variable to access it more easily
         this.intention = this.model.get('intention');
+        this.listenTo(this.intention, 'change:evolvingFunction', this.clearFunctionSegments)
     },
 
     template: ['<script type="text/template" id="item-template">',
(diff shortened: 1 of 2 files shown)


leaf-ui/rappid-extensions/ElementInspector.js
```diff
@@ -68,6 +68,7 @@ var ElementInspector = Backbone.View.extend({
         this.listenTo(this, 'change:intention', this.initSatValueChanged);
         // Saves this.model.get('intention) as a local variable to access it more easily
         this.intention = this.model.get('intention');
+        this.listenTo(this.intention, 'change:evolvingFunction', this.clearFunctionSegments)
```
YesugenB: Add `;` at the end

YesugenB:
Looks good to me! I think you can delete `removeFunction`. I had one question - I noticed that the chart doesn't fully reset to `no value` when you add a user defined function, clear dynamic function type, and then clear evaluation labels. Instead the chart seems to go to `stochastic`. 
Screen Shot 2022-03-11 at 5 26 39 PM

Vbeany:
You don't have to do this right away but I made an issue #643 where you're able to select a function type even when the sat value is no value.

Domily11:
> Looks good to me! I think you can delete `removeFunction`. I had one question - I noticed that the chart doesn't fully reset to `no value` when you add a user defined function, clear dynamic function type, and then clear evaluation labels. Instead the chart seems to go to `stochastic`. Screen Shot 2022-03-11 at 5 26 39 PM

I think that's a separate issue that either has something to do with the chart or userdefined functions, because I actually didn't change much except the assigning things as NT when the buttons are clicked. The rest were just moving what was originally there.

YesugenB (approved):
Looks good then! Thanks for answering my question

Issue #638, opened 2022-02-21
Update Save-Load function to be able to read previous model versions


Domily11:
Save-Load needs to be updated so that models saved before the backbone change can be loaded again

sonorah:
This has been addressed by PR #646

PR #646, opened 2022-04-07
Updated load-save to incorporate older models of BloomingLeaf

sonorah:
Previously, load-save only loaded models that had been saved on the newer version of BloomingLeaf, but now it can also load pre-backbone models. Addresses #638.


Vbeany (approved):
Seems like it works and everything looks good. There's a few console logs on lines 154, 25, 34 to remove though

PR #661, opened 2022-06-14
actors now log merged intentions

anishaj149:
Originally, intentions will log their new actor objects, but actors would not keep track of new intentions. I added a method to update the elements embedded in an actor and added a line in the merge algorithm so that a new actor object will collect the unique IDs of any intentions that the actor version in the other model has but the new one doesn't. (When MMain.DEBUG = true console outputs will show when an intention has been added to a new actor) (Changes in MMain are not important, just for running stuff)

### DIFF
--- a/leaf-analysis/src/merge/MMain.java
+++ b/leaf-analysis/src/merge/MMain.java
@@ -20,7 +20,7 @@ import simulation.BIModelSpecBuilder;
  *
  */
 public class MMain {
-	public final static boolean DEBUG = true;
+	public final static boolean DEBUG = false;
 
 	/**
 	 * This method is responsible to execute all steps to generate the merged model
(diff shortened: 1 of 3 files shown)


Vbeany (approved):
Looks good, muy excelente

leaf-analysis/src/merge/MMain.java
```diff
@@ -20,7 +20,7 @@
  *
  */
 public class MMain {
-	public final static boolean DEBUG = false;
+	public final static boolean DEBUG = true;
```
amgrubb: please set debug back to false.

leaf-analysis/src/merge/MMain.java
```diff
@@ -32,18 +32,18 @@ public static void main(String[] args) {
 		String tPath = "data/timing/";
 		String outPath = "data/mergedModels/";
 		String tracePath = "data/traceability/";
-		String inputFile1 = "";
-		String inputFile2 = "";
-		String timingFile = "";
+		String inputFile1 = "S1-Part4.json";
```
amgrubb: I don't think these should be checked in/hard coded.

leaf-analysis/src/merge/MMain.java
```diff
@@ -32,18 +32,18 @@ public static void main(String[] args) {
...
 		String outputFile = "output.json";
 		
 		try {			
-			if (args.length == 4) {
-				inputFile1 = args[0];
-				inputFile2 = args[1];
-				timingFile = args[2];
-				outputFile = args[3];
-			} else throw new IOException("Tool: Command Line Inputs Incorrect.");
+//			if (args.length == 4) {
```
amgrubb: Why are these commented out?

amgrubb (changes requested):
Please review and update MMain.java.

amgrubb (approved):

PR #679, opened 2022-09-30
Add merge-main changes to develop

krhablutzel:
Now that our RE'22 paper on merge is published, we are ready to add the merge git history to the develop branch of BloomingLeaf. We especially wish for the backend changes to be integrated with develop, so that future development can incorporate our changes.


anishaj149 (changes requested):
I think the majority of this code is well commented and I can follow it -- but also since I wrote a lot of it that may be part of my comfort with it. 
There are a few methods (actually a few of the ones I wrote) that aren't well commented, like the actorLink merging method, that should have stuff added. 
One thing that makes the code understandable is clear naming convention, but some methods I just don't understand what the purpose is or how it accomplishes what it does. E.g. In AbstractElementLink, noLongerEvolves() returns true when isEvolving == false but how do we know it was ever true? 
This is not specific for merge, but it's in the same vein: I think some comments should be added that explain the differences between name, unique ID, and ID because it's not so clear to the reader. And more importantly I don't understand the hierarchy because don't both abstractElementLink and abstractLinkableElement have an id? Maybe should id also be inherited from abstractElement?
Finally I see a good amount of TODOs (decompositionLink.java, ICell.java, MergeAlgorithm.java) and obviously the code works despite adding the code required by the TODOs --- maybe we should look into these and convert them to "future work" or delete them if they aren't necessary.

amgrubb:
How are these changes/updates coming?

krhablutzel:
I agree that we have a ways to go on backend documentation. Currently, the backend has a steep learning curve and assumes a high level of base familiarity. The backend documentation need lends itself to a larger project outside the scope of Merge, though we should hold our Merge additions to higher standards than the existing codebase.

To address the concern on AbstractElementLink, noLongerEvolves() is a method for the analysis backend of BloomingLeaf, not one we introduced with merge. We have a comment in the constructor above explaining that isEvolving represents whether a link has a post relationship type, but I went ahead and added a reminder for clarification by the getter/setter methods.

For the concern on uniqueID vs. id vs. name, I've added more clarifying documentation. The tl;dr is that uniqueID = frontendID, id = backendID, and name = frontendNameInWords. Regarding the object hierarchy, it seems we could move id to AbstractElement; however, id is used mostly on AbstractLinkableElements, which start off with their own backend id, while all AbstractElementLinks start with an id of "NO-ID".

General status update:
- Added comments to more merge methods. Leaving comments on link merging methods to @anishaj149 
- Cleaned up the TODOs.
- Added backend clarification comments to AbstractElementLink and regarding ids, following the suggestions above.

krhablutzel:
As soon as @anishaj149 confirms that her concerns are addressed, then @amgrubb, this should be ready to merge into develop.

anishaj149 (commented):
Looks fab, my concerns have been addressed.

anishaj149 (approved):
Gr8

PR #691, opened 2022-11-14
Bug #689 Fixed

### LINKED ISSUE #689: Font resize does not apply when adding new actors
If you change the font size in your model, then add a new actor element, the actor will be displayed with the default font size:

image

Cyrine-BA:
A newly added actor will now check for the correct font size.


YesugenB (approved):
Looks great!

PR #693, opened 2022-11-15
Font size bugs

### LINKED ISSUE #692: Can't change the font size for elements in Modeling
If you increased/decreased the font size in Modeling then switched to Analysis. The actors will have the correct font size but the rest wouldn't. 
You can also resize the actor's in Modeling, but you can't do that for other elements.

Screen Shot 2022-11-14 at 2 31 33 PM

Cyrine-BA:
Switching from Analysis/Modeling and vice versa will now keep the font size. You can still increase/decrease the font again.
I narrowed down the bug to one code in CSS that was fixing the bug. I think we can just get rid of that line because it doesn't seem to do much either way.


YesugenB (approved):
Looks great!

PR #694, opened 2022-11-25
EVO Final Steps

Cyrine-BA:
EVO is now done and ready to be merged into Develop. These are the main additions/changes: 

1- My Palette
Screenshot 2022-11-25 at 1 47 32 PM

* The user is now able to choose a color from a color pie instead of inputting the Hex Code 
* They HAVE TO click on "Save My Palette" for their selection to be saved. If they wish to cancel, they can just close the window
* The "Choose My Palette" tab is initialized with the default colors. So is the dictionary SelfColorVisDict.

2- Changing Color Selection in Modeling/Analysis:
* EVO Should refresh automatically each time they choose another palette or make new selections and click on "Save My Palette". (This was a bug in the Spring)

3- Color Key
Screenshot 2022-11-25 at 1 59 46 PM
* Help > EVO Color Key displays a dropdown asking, "What Color Key do You want to See?"
*  Once the user clicks on one of the options, the color key is shown. 
* For the personalized palette, the color key should update each time there's a change to the selection.


Cyrine-BA:
Closed by mistake !

anishaj149 (changes requested):
Great works guys, I have some feedback:
1. Save my palette button provides no feedback to user that palette has been saved. It bounces, but I feel like there should be a message or indication that the save has been successful 
2. The evo slider changes when the 'my palette' is changed during analysis mode. You can only turn it on and off in this case, versus the %/Time/State slider
3. When I open the help tab to find the color key, the analysis parameter window disappears. It also does not come back when the I exit out of the color key window. 
4. Is it acceptable that the users can make all the colors in the palette the same; in other words is there no logic to palette coloring?

sonorah:
It's so nice to see this project finally come together!!

I agree with points 1,2,4 that Anisha made.
For 3, I think the analysis parameter window goes away because the intention gets unselected. When I close the help window and click back on the intention it works just fine. 

Here are some other observations I have:
1. It'd be nice if you could change the intention text to white when the user picks a dark color (including black, in which case text is illegible). 
2. In the help window, the box colors and the text aren't aligned perfectly. Not a big deal but it'd be nice if it was more symmetrical. 
Screen Shot 2022-11-29 at 2 09 46 PM

3. Think about re-naming the palettes maybe? What makes the original one default? I remember the whole point of this project was making the tool accessible to international audiences, so why differentiate between the more western one from the rest? Maybe name it the Blue-Red Palette? Just a thought.

Cyrine-BA:
Thank you both for the feedback! 

@anishaj149 we will work on points 1 and 2. For 3, I am not sure what the problem is. According to Sonora, I think we're good, right? For your fourth point, I have to say we haven't thought about that, so thanks for bringing it up. The palette would allow users to pick the same color everywhere but we'll make sure to display an error message if that's the case. If you have any other thoughts about this, please let us know!

@sonorah You're so right about the naming. We will change that right now. The alignment is also an easy fix, but for the text color, we already changed it to white when they picked a dark color (it's shown in the screenshot). Or did you mean the text in the model itself?

sonorah:
Yes, I did mean the text in the model itself @Cyrine-BA.

anishaj149:
Just to clarify my point 3: this block disappears when I open the color key

Screen Shot 2022-11-29 at 10 18 48 AM

@sonorah I'm not sure what you mean by clicking on an intention, clicking on an intention in analysis mode doesn't do anything for me. Is the analysis parameter window connected to intentions? I'm noting that this analysis parameter window remains intact when I open the palettes.

After more investigation, it seems that the window does reappear when I add another 'request' and when I click from analysis to modeling and back to analysis.

I'm not sure if this is a problem or not, but I'm having trouble rationalizing why this *should* happen. Can anyone clarify what is going on?

sonorah:
I think I misunderstood your point then, @anishaj149. I did not try what you demonstrated in the picture.

Cyrine-BA:
@sonorah and  @anishaj149, I have made changes per your requests. Would you have a look at it again?

anishaj149 (changes requested):
Thanks for looking at the suggestions!
1. I don't know exactly if it's a problem. So when Evo is on and the slider is on, for example, Time, any intention will have bars of colors right. When I go into my palette and change the palette to e.g. red-green, the bars of color will change to that palette. But when I change the "my palette" to something, the intentions will change into this: 
Screen Shot 2022-12-05 at 8 43 45 PM
but fortunately when the slider is manipulated again everything looks like its still working.

2. The change where no two values can be the same in the "my palette" resulted in the colorblind palette being invalid in the my palette. I don't know how the logic works but maybe if the colorblind palette is a valid palette then maybe the constraints on the palette should be reworked

sonorah:
The changes I suggested look great! Good job!

amgrubb:
@anishaj149 & @sonorah  Can you resubmit your review and approve it...if it is ready.

Cyrine-BA:
@anishaj149 You were right, it was a bug indeed. I fixed it now, could you have another look? Maybe for the last time? =)

amgrubb:
@Cyrine-BA You deleted Blooming.jar again...

anishaj149 (dismissed):
I think all my concerns have been addressed. Good job guys!

amgrubb:
Replaced by #695.
Too many incorrect files committed on branch.

PR #695, opened 2022-12-08
EVO Final Steps II

### LINKED ISSUE #618: EVO Choose-Your-Own-Colors

amgrubb:
**This pull request replaces #694!**

==================

EVO is now done and ready to be merged into Develop. These are the main additions/changes: 

1- My Palette
Screenshot 2022-11-25 at 1 47 32 PM

* The user is now able to choose a color from a color pie instead of inputting the Hex Code 
* They HAVE TO click on "Save My Palette" for their selection to be saved. If they wish to cancel, they can just close the window
* The "Choose My Palette" tab is initialized with the default colors. So is the dictionary SelfColorVisDict.

2- Changing Color Selection in Modeling/Analysis:
* EVO Should refresh automatically each time they choose another palette or make new selections and click on "Save My Palette". (This was a bug in the Spring)

3- Color Key
Screenshot 2022-11-25 at 1 59 46 PM
* Help > EVO Color Key displays a dropdown asking, "What Color Key do You want to See?"
*  Once the user clicks on one of the options, the color key is shown. 
* For the personalized palette, the color key should update each time there's a change to the selection.

Fixes #618


amgrubb:
When the user edits `My Palette` then selects one of the other color systems, `My Palette` is changed to the new selection. Was this the intended use? Essentially, `My Palette` is just the current palette.

Question: should we add another button, called `Edit My Palette` that allows the user to change and select the *customized* palette, then the first item `My Palette` would allow the user to select (but not change) the custom palette and toggle between others?

Thoughts?

Cyrine-BA:
I realized that My Palette is essentially the current one, but I was thinking it would allow the user to select the pre-made palette that aligns the most with what they want. Then they can change some of it but not necessarily all of it.

But the "edit My Palette" makes more sense. I will work on that. @mvarnum and @bobiamack let me know if you're free too.

Cyrine-BA:
@amgrubb The bug is fixed, and I added the "Edit My Palette" option. Let me know if there's anything else you'd like me to change.

mvarnum (dismissed):

violetbi (commented):
Everything looks perfect to me. There are enough comments and no redundant codes.

violetbi (commented):
Everything looks perfect to me. There are enough comments and no redundant codes.

violetbi (approved):

PR #706, opened 2023-02-16
Last sat val of UD function segment no longer resets 👍

### LINKED ISSUE #635: Satisfaction value in Function segment view turns blank in UD
When in UD, if a user chooses a satisfaction value in the funcSeg view, click on paper, and then click on the intention again, the satisfaction value first turns to blank, then after the User adds another function segment, the autofill overwrites the blank.

image
image
image

### LINKED ISSUE #700: Element Inspector: UD Func - Save Last Seg
On `develop`,  if I create a UD function, the set value for the last function is not saved. In the following example, I set it to be satisfied, but when I navigate off and then back on, it is None.

Before:
Screen Shot 2023-01-17 at 9 16 35 AM

After:
Screen Shot 2023-01-17 at 9 16 46 AM

mvarnum:


stardess:
We think that our 700 bug fix also got rid of the issue in 635. 😁

Vbeany:
Everything looks good with me in terms of the last sat val being saved and editable for the most part.

 Although one thing that I noticed was that whenever you set the function type to either decrease or increase and select within the element inspector for the first time to change the Satisfaction type you get 4 options and then when you click on to the paper and back on there are 5 options now. Idk if that added option is supposed to be an option or not.
Screenshot 2023-03-01 at 11 48 58 AM
Screenshot 2023-03-01 at 11 49 11 AM

Secondly I'm not sure if this is supposed to be like this either but whenever you select the function type as constant and select a satisfaction type. You can still change and edit it but when you click on the paper and return to the element inspector the satisfaction is saved and unchangeable? 

Screenshot 2023-03-01 at 12 04 45 PM
Screenshot 2023-03-01 at 12 06 24 PM

Vbeany (changes requested):
I left comments about some things I was unsure we had wanted

stardess:
Megan, Bobi and I just looked over the bug again and it should be fixed by now. Please review and let us know your comments

Vbeany (dismissed):
Everything looks good to me! 👍

emruckle (approved):
Tested the changes and everything is working (creating models, saving models, loading models)

mollyd13 (approved):
tested the changes and everything looks good!

PR #707, opened 2023-02-22
Reset element inspector after clearing evaluation labels 

### LINKED ISSUE #643: Several function types pop up even when the sat value is none
image

When I clear dynamic function labels then clear evaluation labels even when sat value is no value I can now select multiple function types when I should only be able to select the stochastic function only.

### LINKED ISSUE #644: Chart doesn't reset properly after adding UD function
The chart doesn't fully reset to no value when you add a user defined function, clear dynamic function type, and then clear evaluation labels. Instead the chart seems to go to stochastic
157981561-2db3b04c-a893-48c1-ac30-24d306407b4b
.

mvarnum:
Fixed #644 & #643

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -179,11 +179,8 @@ $('#btn-clear-flabel').on('click', function () {
             $(".function-type").val('NT');
             cell.attr(".funcvalue/text", "");
 
-            if ($('.inspector-views').length != 0) {
-                // Rerender elementInspector for clearing Dynamic Labels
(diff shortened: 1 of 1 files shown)


Vbeany (approved):
Issue #644 looks resolved!
Issue #643 when clearing dynamic function labels then clearing evaluation label everything is reset which seems right! One comment is we're not sure about when you first select the Initial sat val to be satisfied for ex. and then choose Stochastic-Constant as the function type, the Initial Sat val is set to (no value) but you can still choose other choices besides stochastic. I think an issue should be made for this but besides that everything looks good.

PR #710, opened 2023-03-09
Presence Conditions Front-End

amgrubb:


stardess:
All the files changed look good overall; the variables have good names, and the code is well commented and easy to follow. However, there was one file that seems to have an error. In the file BackboneModelsIntention.js the refEvidencePair is set to equal to the int 0000. The code editor has a redline underline on that line signifying that the value's type should be a string and not an int. Once this is revised, the code should be good to go.

amgrubb:
@kkung9 Do you think you might have a few minutes to take the _quickest of looks_ at this?

stardess (commented):
All the files changed look good overall; the variables have good names, and the code is well commented and easy to follow. However, there was one file that seems to have an error. In the file BackboneModelsIntention.js the refEvidencePair is set to equal to the int 0000. The code editor has a redline underline on that line signifying that the value's type should be a string and not an int. Once this is revised, the code should be good to go.

leaf-ui/rappid-extensions/ActorInspector.js
```diff
@@ -93,4 +142,209 @@ var ActorInspector = Backbone.View.extend({
...
+    },
+    template: [
+        '<script type="text/template" id="item-template">',
+        '<div class="container">',
+            '<div class="slider-track">',
+            '</div>',
+            '<input type="range" min="0" max=<%= graph.get("maxAbsTime") %> value="0", id="slider-1" oninput="slideOne()">',
+            '<input type="range" min="0" max=<%= graph.get("maxAbsTime") %> value="<%= Math.round(graph.get("maxAbsTime"))%>", id="slider-2" oninput="slideTwo()">',
+        '</div>',
+        '<label for="range1">Available: ',
```
violetbi, line 300: Currently, when interval like 0-50 is flipped, this will display 0-0, 51-100, although the value is stored correctly (exclusion 0-50). However, do we want to fix this before merging into develop. or should we open an issue?

leaf-ui/rappid-extensions/ActorInspector.js
```diff
@@ -7,23 +9,69 @@ var ActorInspector = Backbone.View.extend({
...
+                '<textarea class="cell-attrs-text" maxlength=100> <%= actorName %> </textarea>',
+        // Actor type
+                '<label> Actor type </label>',
+                '<select id="actor-type-ID" class="actor-type">',
+                    '<option value=A <%if (type === "A")  { %> selected <%} %>> Actor </option>',
+                    '<option value=G <%if (type === "G")  { %> selected <%} %>> Agent </option>',
+                    '<option value=R <%if (type === "R")  { %> selected <%} %>> Role </option>',
+                '</select>',
+        // Hide Actor checkbox and label
+                '<input type="checkbox" id="actor-hidden" name="hidden" value="true" style="float: left; margin-top: 25px; margin-left: 65px;"><label for="actor-hidden" style="float: left; margin-top: 25px; margin-left: 10px; margin-bottom:40px">Hide Actor</label>',
```
violetbi, line 24: Currently click on it, embedded elements are not hidden.

stardess (dismissed):
I think the code could be merged now.👍🏼

leaf-ui/js/onFunctions.js
```diff
@@ -753,6 +754,15 @@ paper.on("link:options", function (cell) {
                 selectResult.set('selected', false);
             }
 
+            var elements = SliderObj.getIntentionsAndActorsView();
```
violetbi, line 757: I'm wondering what are these codes for? I don't think it's being used in this method or I'm just missing something

violetbi (commented):
There is still a lingering problem with exclusion interval storage. If we designate one's interval from 0-100 to 0-50, then change the maximum time point from 100 to 110. then available would be 0-51, 101-110 and the flip button would gone in this case. One solution might be makes the maximum absolute time fixed once the user move the interval slider. This can also avoid case when the user is trying to set a maximum absolute time that is smaller than the current adjusted interval.
We can leave it as an issue if Alicia doesn't think it's a serious issue.

Except several issues, it's looking in good shape! The way fields are added and updated won't really break develop branch.

leaf-ui/rappid-extensions/ActorInspector.js
violetbi: We can leave it as an issue if Alicia doesn't think it's a serious issue.

leaf-ui/rappid-extensions/ActorInspector.js
violetbi: We can leave it as an issue if Alicia doesn't think it's a serious issue.

leaf-ui/rappid-extensions/ActorInspector.js
kkung9: fixed in commit 258c629

leaf-ui/rappid-extensions/ActorInspector.js
kkung9: fixed in commit a5595c9

leaf-ui/js/onFunctions.js
kkung9: I think this was to make sure all elements show up in modeling mode, even if they had been hidden at the most recent time point due to their presence condition interval

kkung9: initial comment re: exclusion interval storage when increasing max abs time was fixed in commit 8d1a57e

leaf-ui/rappid-extensions/ActorInspector.js
violetbi: Links are not hidden when embedded intentions are. It might be because links are not part of cellView.

kkung9: fixed with 6bbd4cc

violetbi:
Pasted Graphic 4
found another issue: if you drag both ends of the interval and flip interval, the flip interval button will disappear when you re-click on the actor.

mollyd13 (commented):
Code is well commented and well formatted. Next state and analysis are working as expected, and the sliders look good! Only one question: When you make a change to the slider interval, the interval is highlighted in blue. However, when you click off of the intention and then back on, the interval is no longer blue. If this is how it's supposed to function then all is good! But if this is not supposed to happen, consider looking into this. 

Slider immediately after changing the interval:
Screenshot 2023-10-02 at 4 52 58 PM


Slider after clicking off and returning to the intention: 
Screenshot 2023-10-02 at 4 53 07 PM

kkung9 (approved):

amgrubb:
@mollyd13 Can you also approve this request and ensure that your requested changes were corrected.

mollyd13 (approved):
Code looks good, and the chosen interval now stays blue when you click off and back on to an intention.

PR #711, opened 2023-05-17
Ev0 deleted changes

cbenayed:
This branch builds on the last PR for EVO, but some last changes were lost and/or not fully functional
- a confirmation message to tell the user their palette options have been saved
- show an error message when the chosen satisfied and denied values are the same 
Additional features now include:
- a cancel button to cancel all new changes to my palette => it restores it to what it was when "save changes" was last clicked on

- when the user picks the same color (same hex code) for satisfied and denied they not only get the error message but the "create my palette tab" remains open and the values for satisfied and denied don't change until they pick different ones as shown in these three pictures

1- Make changes (picked the same color for satisfied and denied on purpose)
Screenshot 2023-05-17 at 3 59 14 PM

2-save
Screenshot 2023-05-17 at 3 59 26 PM

3-click on ok ( satisfied and denied go back to what they were before automatically and the color key remains untouched)
Screenshot 2023-05-17 at 3 59 34 PM

### DIFF
--- a/leaf-ui/css/style.css
+++ b/leaf-ui/css/style.css
@@ -688,3 +688,17 @@ padding: 0px 20px 0px 0px;
   padding: 5px;
   text-align: center;
 }
+
+/** buttons for the palette customization */
+.button-container {
+  display: flex;
+  justify-content: space-between;
+}
+
+.top-left-button {
+  align-self: flex-start;
+}
+
+.top-right-button {
+  align-self: flex-start;
+}
--- a/leaf-ui/index.html
+++ b/leaf-ui/index.html
@@ -134,12 +134,9 @@
             <div id="page-mask"></div>
                     <div class="popup_content">
                     
-                        <div class="popup_header">
-
-                            <span id="submit-color" class="close" onclick=" $('#color-input').css('display', 'none');">&times;</span>
-
+                        <div class="popup_header"> 
                             <h2>Create My Palette</h2>
-                        </div>
+                        </div>                            
                         <div class="popup_body">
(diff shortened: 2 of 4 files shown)


mollyd13 (commented):
The changes you made look good, but it still allows you to have multiple values with the same exact color as long as satisfied and denied are different. If this was your intention, it works, but if that's something that shouldn't be allowed, more changes should be made.

emruckle (commented):
The "save" confirmation message, error when satisfied and denied values are the same, and the cancel button all work correctly. But you are able to select the same color for either satisfied and every other satisfaction value in the palette besides denied without getting an error. Same goes for denied. I've attached an example that saves without an error. Is this something we want to address?
Screen Shot 2023-05-17 at 3 21 43 PM

cbenayed:
@mollyd13 @emruckle would you please have another look? The customization should only work if the chosen color for none, satisfied, denied, and FF are all distinct

emruckle (dismissed):
Checked all combinations of denied, none, satisfied and ff. The user cannot select the same value for any two of those states. The pop up successfully appears each time I attempt to assign two (or more) states the same color. Also checked simulate single path. "My palette" is working correctly in conjunction with simulate single path.

mollyd13 (approved):

mollyd13 (dismissed):
Code is well commented and well formatted. Tested colorblind mode and loading/saving files, everything works as expected.

mollyd13 (approved):
tested simulate single path and next state with evo, everything looks good.

emruckle (approved):
Code is clear and well commented. Satisfied and denied in color blind palette correspond to the correct colors in red-blue palette. Tested simulate path and explore states and everything works.

PR #712, opened 2023-05-23
Output tag bug fixed

mollyd13:
Added current attribute to the output json of merged models.


mollyd13:
Tested simulate path, next state, and saving/loading models with current tag change. Everything is working as expected.

Issue #452, opened 2021-06-24
Modifying updateType in the ActorInspector


Domily11:
Use the code in updateType and find a method to directly listen to the model instead of taking the information from the interface, the current problem is that actorType is within the actor parameter (actorBBM) but is dealing with attributes from the basic.Actor model

Emkung:
issue fixed

PR #715, opened 2023-06-13
fixes the issue

Emkung:
I fixed the issue #452


czouloumian (changes requested):
There is an Uncaught TypeError on evoObjects, line 514, when dragging an element. The code would also need some more comments

PR #716, opened 2023-06-19
Fix clear of evaluation evaluation and evidence pairs

### LINKED ISSUE #708: Clear evaluation label doesn't clear the sat val unless clicked twice or intention is selected
When an intention has both a function type and sat val and when you first clear the dynamic function type and then clear the evaluation label you have to select the evaluation label sometimes more than twice to have the sat val disappear or if click on the intention the sat val will disappear. This only occurs when you are not clicking on a specific intention but clicking on the paper.
Screenshot 2023-03-01 at 3 49 34 PM

Emkung:
fixed the issue in which an error pops up when an intention was set with a sat value and a function type and the user clears the evolving function labels and evaluation labels. 

We fixed it by not displaying the chart after clearing.

Fixed #708


czouloumian (approved):
The code is clear and well commented, and the error is solved.
Loading, saving, simulate single path, EVO and next state still work.

PR #717, opened 2023-06-20
Fix disappearing slider when switching color palettes

### LINKED ISSUE #663: The time point slider disappears when you select a color palette but you're in EVO "off" mode
Screen Shot 2022-06-21 at 9 53 56 AM

Emkung:
fixed the issue in which the slider disappears when switching color palettes by updating the variable select results when simulate path is clicked

Fixed #663


czouloumian (approved):
The error is solved and the code is clear. Loading, saving, simulate single path, EVO and explore next state still work.

PR #718, opened 2023-06-20
Fix analysis results showing up from an old model after loading in a different model

### LINKED ISSUE #669: Old results aren't cleared when a different model is loaded.
Results from old/original model are not cleared in the sidebar when you load a model and go into analysis mode

kkung9:
Fixes Issue 669, in which Results from an old model would show up in Analysis after loading in a new model.
This issue was fixed by resetting configCollection when loading a new model.
The solution was tested by creating a model and simulating paths, then loading saved models both with and without saved results. 

Also found and fixed a bug in which loaded models do not have a slider element during simulation.

Fixed #669

### DIFF
--- a/leaf-ui/js/loadSaveFunctions.js
+++ b/leaf-ui/js/loadSaveFunctions.js
@@ -64,6 +64,8 @@ function loadFromObject(obj) {
 	// If the object contains configCollection, create configCollection fields from JSON
 	if (obj.configCollection != undefined) {
 		loadConfig(obj.configCollection)
+	} else {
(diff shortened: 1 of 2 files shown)


emruckle (approved):
Code is clear and well commented. Fixes analysis results issue. Tested by creating models, conducting analysis and then saving the analysis and results. When we load that back in, the correct analysis loads.

mollyd13 (approved):
Code is well formatted and clear. Tested simulate path and saving analysis requests. The issue seems to be fixed!

PR #720, opened 2023-06-21
Disable Explore Next States button on last state

### LINKED ISSUE #687: In Next State, turn "Explore Next States" gray if on last state
If the user is viewing the last state in Next State, we want the "Explore Next States" button to turn gray and inactive. This will replace the current popup which signals the last state.

kkung9:
Fixes Issue #687 by disabling the "Explore Next States" button when viewing the last state. The button is no longer clickable and is visually greyed out. 

This can be tested by increasing the number of relative time points in the simulation, then using the Explore States button and clicking through to the last state.

Fixed #687

### DIFF
--- a/leaf-ui/css/analysis_html.css
+++ b/leaf-ui/css/analysis_html.css
@@ -248,6 +248,12 @@ body {
   transform: translateY(2px);
 }
 
+.inspector-btn.disabled {
+  opacity: .75;
+  box-shadow: none;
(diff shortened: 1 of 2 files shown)


stardess (approved):
The code looks fine. It is understandable and well commented. We also tested loading and saving a model, simulate single path, explore next states and EVO and they all work fine.

PR #721, opened 2023-06-21
Displays time point in Next State

### LINKED ISSUE #686: In Next State, display which # state is currently shown
In the Next State window, we want to display in the interface which number state is currently shown.

kkung9:
Fixes Issue #686 by adding text in the left panel that displays which time point is being currently modified in the panel. This can be tested by exploring Next States.

Fixed #686

### DIFF
--- a/leaf-ui/analysis.html
+++ b/leaf-ui/analysis.html
@@ -110,6 +110,7 @@
 
     <div id="navigation" class="left-panel" style="background-color: #333;">
         <br>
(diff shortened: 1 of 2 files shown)


emruckle (approved):
Code is clear. The explore next states pop up displays the time point being modified in explore next states. When you move to the next time point within the pop up, the time point number updates correctly. Doesn't break explore next states, simulate single path, loading models and saving models.

mollyd13 (approved):
Code looks good. Tested explore states and the timepoint in the popup correctly corresponds with the selected timepoint in analysis. Explore next states popup updates the timepoint as expected.

PR #722, opened 2023-06-26
Fix slider disappearing in EVO off mode

### LINKED ISSUE #663: The time point slider disappears when you select a color palette but you're in EVO "off" mode

kkung9:
Fixes the issue in which the slider disappeared if the color palette was changed before moving the EVO slider. Fixed by setting the selectResult variable when the new path is simulated.

Fixed #663

### DIFF
--- a/leaf-ui/js/onFunctions.js
+++ b/leaf-ui/js/onFunctions.js
@@ -1080,7 +1080,6 @@ paper.on("link:options", function (cell) {
      */
     document.getElementById("colorResetAnalysis").oninput = function () { // Changes slider mode and refreshes
         var selectConfig;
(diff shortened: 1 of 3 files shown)


emruckle (approved):
Code is clear and well commented. Fixes slider issue in EVO off mode. Tested simulate single path and explore next states. Also tested moving between modeling and analysis with EVO on and off, and toggling between EVO states in analysis.

mollyd13 (approved):
Code looks great! Tested changing the color palette and the slider no longer disappears after turning on evo. Simulate path still works as expected, and next state is not broken.

PR #723, opened 2023-07-06
Completed Merge Interface

mollyd13:
Updated the timeline popup to include draggable elements. The user can now set timepoints to happen simultaneously.


kkung9 (commented):
The code is functional and successfully merges two models, including when there are repeating functions and equal time points within a model. The changes do not break simulate path, user defined functions, saving and loading models and results, or next state. A minor issue is that satisfaction values show up as numbers rather than symbols after merging until an intention is clicked on.

A few parts of the code to clean up:
- Remove console log statements in several files
- Remove large commented out dead code in backendComm
- Change the path in app.js back to judySmith

PR #724, opened 2023-07-07
Next State: Update view when no state is available.

### LINKED ISSUE #666: add warnings so users can't have no solutions in next state
currently users can add as many filters as they want in next-state until there are no solutions, which causes EVO % mode to break (and a solution is still shown on the graph, which seems inaccurate) 
Screen Shot 2022-06-22 at 10 21 39 AM

Emkung:
fixes the issue, in which next state breaks when there are no more states to navigate, by adding a pop-up warning, removing the last selected filter, and rerendering the navigation sidebar.
 
Also, I fixed some minor bugs, such as state 0 is not reachable when typed in and errors popping up when there are no more solutions due to some functions still running

Fixed #666

### DIFF
--- a/leaf-ui/js/next-state.js
+++ b/leaf-ui/js/next-state.js
@@ -18,6 +18,8 @@
     // An array of all of the solutions and every element is another array 
     // with all of the refEvidencePairs for the intentions at that solution
     var allSolutionArray = [];
+    var lastAllSolutionArray =[]; //variable to save the last solution array with solutions
+    var lastFilter = ""; //variable to save the name of the last filter clicked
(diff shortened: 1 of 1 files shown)


kkung9 (commented):
The changes address the issue and do not seem to create bugs in Next State or simulation.

The text of the popup "No more states to display" is a little confusing to me, something like "Cannot apply filter; no possible states with that combination of filters" would make the warning more clear in my opinion.
The variables "lastAllSolutionArray" and "lastFilter" have a global scope.

kkung9 (approved):

PR #725, opened 2023-07-07
Summer23 bug fixes

kkung9:


emruckle (dismissed):
Code is clear and well commented. Addresses issues with clearing evolving function labels, changing color palettes, and selecting a state in next state when there are a high number of possible states. Does not break simulation, setting evolving functions, or choosing next states

mollyd13 (changes requested):
Code looks good and has sufficient comments! Testing confirms that bugs #708, #669, #667, #687, #686, and #663 are fixed successfully. However, issue #666 seems to not be fixed. When you apply a filter that leads to having 0 states, a popup appears telling you it is not allowed and the filter is unselected, but the number of states remains at 0 you can no longer save and close next state.

mollyd13 (approved):

amgrubb:
Replaced by #726.

Issue #667, opened 2022-06-22
Page selection for Next State disappears with large numbers


YesugenB:
<img width="208" alt="Screen Shot 2022-06-22 at 11 42 46 AM" src="https://user-images.githubusercontent.com/59470488/175074301-ce86c16a-1be9-4d95-b80e-132445da64bd.png">

mvarnum:
It looks like the page selection is still there, you just have to scroll over:

<img width="369" alt="image" src="https://user-images.githubusercontent.com/55264239/216852892-0c3b8abb-28c6-4741-9dbf-93f4685f3e01.png">

PR #726, opened 2023-10-18
Select sum23 bug fixes

amgrubb:
Replaces PR #725 

@kkung9 @mollyd13  Please reconfirm via testing that bugs #708, #667, #687, #686, and #663 are fixed successfully. 

@YesugenB Please test and check next state changes, and EVO slider appearing disappearing correctly.

### DIFF
--- a/leaf-ui/js/next-state.js
+++ b/leaf-ui/js/next-state.js
@@ -210,6 +210,9 @@
         updateNodesValues(currentPage);
 
         renderEVO();
+
+        document.getElementById("thisState").textContent = myInputJSObject.results.get('timePointPath').length;
+        document.getElementById("totalStates").textContent = myInputJSObject.results.totalNumTimePoints.length - 1;
     }
 
     /**
--- a/leaf-ui/js/next-state.js
+++ b/leaf-ui/js/next-state.js
@@ -240,34 +240,29 @@
      * The number of the page that is selected in the next State window 
      */
     function updatePagination(currentPage) {
+        var currentDigits = currentPage.toString().length;
         var pagination = document.getElementById("pagination");
         var nextSteps_array_size = allSolutionArray.length;
         if (nextSteps_array_size > 6) {
             renderPreviousBtn(pagination, currentPage);
             if (currentPage - 3 < 0) {
-                for (var i = 0; i < 6; i++) {
+                for (var i = 0; i < 7; i++) {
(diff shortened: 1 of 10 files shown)


YesugenB (commented):
EVO slider does not work properly in Analysis Mode when a path has been simulated. 

1. When EVO is turned off and you go into Analysis mode to simulate a path and turn on EVO into % or Time mode, the slider does not disappear. 
2. When EVO is turned on and you go into Analysis mode to simulate a path, the default is EVO % mode and the slider does not appear, which is good. However, when you go to EVO State Mode, the slider does not appear when it should. But when you turn EVO off, the slider appears. Then it does not disappear when you go back to Time or % mode.

leaf-ui/js/next-state.js
```diff
@@ -197,6 +197,14 @@
         var pagination = document.getElementById("pagination");
         var num_states_lbl = document.getElementById("num_states_lbl");
         var currentPageIn = document.getElementById("currentPage");
+
+        // disable Explore Next States button if last time point
+        var exploreNextStatesBtn = document.getElementById("exploreNextStates");
```
YesugenB, line 202: If the last time point is selected, the Explore Next States button is disabled. However, it is not re-enabled when a different, earlier time point is selected. This remains even when you load a new Explore States window from the same time point.
Screen Shot 2023-10-19 at 1 08 23 PM

kkung9 (commented):
Specified issues are addressed. The EVO slider issue mentioned above is fixed in the Presence Conditions branch but not in this branch.

mollyd13:
Issues #708, #667, #687, #686, and #663 seem to be resolved. See Yesugen's comment for EVO slider and next state changes.

amgrubb:
I pulled in the changes from develop, which adds presence conditions. Let's re-test everything and make sure the bugs are fixed.

leaf-ui/js/next-state.js
mollyd13: @YesugenB We are having trouble recreating this issue. Could you clarify what you mean by "selecting an earlier timepoint." Do you mean closing the next state tab and reopening a new one at an earlier timepoint from the analysis tab? Or is there a way to get to an earlier timepoint from the next state window?

mollyd13 (approved):
Bugs are still fixed, and slider works as expected.

PR #732, opened 2024-03-04
QA branch 

### LINKED ISSUE #648: Disable options on IntentionFilter
-Disable invalid dropdown menu options on IntentionFilter based on satvalue (this causes the error on line 148 in nextstate.js file), -Add a warning/disable for when user tries to add duplicate filters to the same intention (right now the user can't, but window doesn't explain why)

### LINKED ISSUE #651: Better color choices for different palettes
For the yellow-purple color palette, maybe change the text color or the shade of the yellows. Once the simulated single path has been run, it's hard to see the text. Also the lime green is pretty harsh on the eyes

### LINKED ISSUE #666: add warnings so users can't have no solutions in next state

### LINKED ISSUE #728: Can set absolute time points outside of max absolute time
To reproduce:
Under View Assignments List, type 101 into the Absolute Time Points box
Add an intention and try to Simulate Path
Throws a RunTime Exception

### LINKED ISSUE #729: Can set absolute intention and absolute relationship assignments to be negative
To reproduce:
Give an intention an initial satisfaction value, such as denied, and a function type, such as denied satisfied
In View Assignments List, under absolute intention assignments, enter something negative for assigned time
Try to simulate path, error appears that says that model is not solvable

### LINKED ISSUE #730: Color Key Bug
When you try to visualize the color key, if you leave the window open and click help > EVO color Key > choose one, all the colors turn white. 
Screenshot 2024-02-23 at 9 54 45 AM

### LINKED ISSUE #731: EVO text border bug
Elements with certain satisfaction values/color keys should have white text in EVO. If they are added to the page when EVO is already on, the text has a black border that makes it hard to read.

cbenayed:
In this branch we: 

- created 2 new EVO color palettes: traffic-light and pastel 
- updates all color palettes to meet the 3:1 contrast ratio as per WWW standards
- updated the isDark() function to fix bug #651 
- Fixed EVO bugs #730 and #731 
- Fixes bugs #728 and #729


cbenayed:
**ADDITION: BUG #666 and #648 FIXES**

1- User can only select intention filters that are part of the general solution. Foe example, in the picture below the % mode tells us that the intention could either be partially denied or fully denied and those are the only two filter options available.
Screenshot 2024-03-11 at 12 01 04 PM

2- (no value) is no longer an option in the dropdown. If user selects a value by mistake, they can either not click "Apply" or just remove the filter from the "Applied Intention Filters" table.

3- Applying the same filter to the same intention generates the error:
Screenshot 2024-03-11 at 12 01 15 PM

mollyd13 (dismissed):
New palettes look good! For the pastel palette, when you assign a satisfaction value to an intention and then put the satisfaction val back to No Value, the font color remains white which is very hard to see on a yellow background: 
Screenshot 2024-03-13 at 1 21 33 AM
Bugs #651, #730, #731, #728, and #729, #666, #648 seem to be all fixed :)

cbenayed:
@mollyd13 would you check again? It should be fixed now!

kkung9 (approved):
The new color palettes look good and fixes the tagged bugs.
A few related issues:
- Bug #739 is fixed such that the absolute time point can no longer be set to be negative, but the absolute time point can still be set outside of the max absolute time, which creates a runtime error in analysis
- the text color in % and time modes depends on whether evo was in off (text always black) or state (text color depends on background color) mode more recently, which seems like an inconsistency

mollyd13 (approved):
Font color issue is fixed, and bugs are still fixed.

PR #735, opened 2024-04-16
Added relative constraints to next state.

amgrubb:
The bug is that if two evolving function have a `<` constraint between their time points, this was not consider in the next state calculations.

_Note: after this code is accepted, the `Blooming.jar` needs to be repacked and updated. I would prefer to keep that a separate PR for more review._


YesugenB (approved):
Looks good. Next State saves the correct path.

PR #743, opened 2025-07-18
Methodology paper

kkung9:
Changes the format of tutorial content from CSV to XML
Updates tutorial text and videos with structure according to methodology paper
Adds root/leaf highlighting option in the Edit menu


amgrubb:
I updated how to buttons work, to not overload the buttons.

amgrubb (approved):
`errorDection.js` seems like the wrong file to put `rootLeafSearch()`, but `onFunctions.js` is similarly too large.

kkung9:
I put it there because that's where the other highlighting occurs, but I agree that it doesn't fit into that category.

amgrubb:
Agreed, that's why I left it there.

